PR #16, opened 2020-04-06
updating image tags for 0.21.1 release

markmark206:
* updating the tags of temporalio/* images for `0.21.1` release,
 * removing unused (for now) section for the web component.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -8,7 +8,7 @@ debug: false
 server:
   image:
     repository: temporalio/server
-    tag: 0.20.0
+    tag: 0.21.1
     pullPolicy: IfNotPresent
 
   # Global default settings (can be overridden per service)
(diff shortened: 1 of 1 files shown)


values.yaml
```diff
@@ -8,7 +8,7 @@ debug: false
 server:
   image:
     repository: temporalio/server
-    tag: 0.20.0
+    tag: 0.21.1
```
shawnhathaway, line 11: Be aware of https://github.com/temporalio/temporal/pull/277, I think you may need some changes here to plumb pod IP to `TEMPORAL_BROADACAST_ADDRESS` env var

markmark206: perfect, thank you for the heads up, will keep an eye on it / make changes as needed.

PR #30, opened 2020-05-28
Add Temporal-web and optional Elasticsearch

pauldemarco:
Adds temporal-web.
Adds `elasticsearch.enabled` option to values.

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -30,6 +30,7 @@ dependencies:
   - name: elasticsearch
     repository: https://helm.elastic.co
     version: 7.6.2
+    condition: elasticsearch.enabled
   - name: kafka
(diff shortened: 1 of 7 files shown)


markmark206:
This is great, thank you so much for this contribution.

As a heads up – we will likely change `elasticsearch.enabled` default to `true` in a bit, while we are putting together a smaller, onebox-friendlier user experience, in addition to the current "everything is included" experience.

Of course, in the meantime, you can always override this in your local `values.yml` (thanks to your change, this value is now explicit;) or in the command line.

```
~/temporal-helm-charts $ helm install temporal-test --set elasticsearch.enabled=false .
```

Thank you for your contribution!

PR #32, opened 2020-05-29
Install 3 replicas of temporal services

markmark206:
Install 3 replicas of each of temporal's services (except web).


markmark206:
leaving the default of 1. our test scenarios that require other configuration are addressed by passing the appropriate replica count to `helm install ... --set ...`

PR #33, opened 2020-06-01
fix indent error

ruanwz:
fix indent error: 
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: [ValidationError(Ingress): unknown field "rules" in io.k8s.api.networking.v1beta1.Ingress, ValidationError(Ingress): unknown field "tls" in io.k8s.api.networking.v1beta1.Ingress]

### DIFF
--- a/templates/web-ingress.yaml
+++ b/templates/web-ingress.yaml
@@ -21,24 +21,24 @@ annotations:
     {{- end }}
 spec:
   {{- if .Values.web.ingress.tls }}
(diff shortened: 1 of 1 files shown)


underrun:
Thanks for the contribution. As a result of our new approach to helm charts this won't be relevant - we will make sure that indentation is correct in the future though. See here for more details. https://github.com/temporalio/proposals/issues/42

PR #39, opened 2020-06-25
Add the option to install without kafka / grafana / prometheus

markmark206:
Add the option to install without kafka / grafana / prometheus (e. g. `helm install --set grafana.enabled=false temporaltest .`).

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -53,7 +53,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.1.1
+version: 0.2.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -30,6 +30,7 @@ dependencies:
   - name: elasticsearch
     repository: https://helm.elastic.co
     version: 7.6.2
+    condition: elasticsearch.enabled
   - name: kafka
     repository: https://charts.bitnami.com/bitnami
     version: 7.2.9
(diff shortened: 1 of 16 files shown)


README.md
```diff
@@ -26,14 +26,23 @@ Temporal can be configured to run with a couple of database choices.
...
 ```bash
 ~/temporal-helm$ helm install temporaltest . --timeout 900s
 ```
 
+To use your own instance of ElasticSearch, MySQL, or Cassandra, please read the "Bring Your Own" sections below.
+
+Other components can be omitted by setting their corresponding 'enable' flag to `false`, e. g.
+
+```bash
+~/temporal-helm$ helm install --set grafana.enabled=false temporaltest . --timeout 900s
```
alexshtin, line 42: This looks slightly confusing (at lease to me). `enabled=false` means "disable". One can get impression that after this command his graphana cluster will be disabled. Or at least grphana features in Temporal will be disabled. `graphana.install` looks like a better name for me.

alexshtin: It might be applicable to other PRs also.

markmark206: oh that's a good point. i think this is a bit of a convention in in helm world to use `enabled` to indicate that the thing is omitted from installation. i'll think on this a bit, to see if there is a better way.

Chart.yaml
```diff
@@ -27,16 +27,19 @@ dependencies:
...
     repository: https://charts.bitnami.com/bitnami
     version: 7.2.9
+    condition: kafka.enabled
   - name: grafana
     repository: https://kubernetes-charts.storage.googleapis.com
     version: 5.0.10
+    condition: grafana.enabled
 
 # A chart can be either an 'application' or a 'library' chart.
 #
```
kfoxb, line 45: I think it'd be nice to update to version in this file with this change. To 0.2.0 probably :-)

PR #41, opened 2020-07-07
Fix yaml indents in web-deployment

yvanin:
Affinity and tolerations belong to `PodSpec` instead of `PodTemplate`, other yaml files got it right: https://github.com/temporalio/temporal-helm-charts/blob/master/templates/server-deployment.yaml#L119-L126


underrun:
Thanks for the contribution. As a result of our new approach to helm charts this won't be relevant - we will make sure that indentation is correct in the future though. See here for more details. temporalio/proposals#42

PR #60, opened 2020-08-19
Using dynamic config in helm

vitarb:
This diff presents helm configuration change that allows users to provide and modify dynamic config properties for the k8s cluster.

In order to modify property values user would simply go to `values.yaml` file and change property value to their liking and then they could run `upgrade` command in order to propagate changes.

New properties would have to be added to the `templates/server-configmap.yaml` into dynamic_config.yaml section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,15 +299,12 @@ Forwarding from [::1]:8081 -> 3000
 ### Updating dynamic config
 By default dynamic config is empty, if you want to override some properties for your cluster, you should:
 1. Create a yaml file with your config (for example dc.yaml).
-2. Populate it with some values under server.dynamicConfig prefix, for example:
+2. Populate it with some values under server.dynamicConfig prefix (use the sample provided at `values/values.dynamic_config.yaml` as a starting point)
+3. Install or upgrade your helm configuration:
+```bash
+$ helm install -f values/values.dynamic_config.yaml temporaltest . --timeout 900s
 ```
-server:
-  dynamicConfig:
-    frontend.enableClientVersionCheck:
-    - value: true
-      constraints: {}
-```
-3. Install or upgrade your helm configuration passing `-f dc.yaml` as a parameter to helm.
+
 
 ## Uninstalling
 
--- a/README.md
+++ b/README.md
@@ -300,11 +300,14 @@ Forwarding from [::1]:8081 -> 3000
 By default dynamic config is empty, if you want to override some properties for your cluster, you should:
 1. Create a yaml file with your config (for example dc.yaml).
 2. Populate it with some values under server.dynamicConfig prefix (use the sample provided at `values/values.dynamic_config.yaml` as a starting point)
-3. Install or upgrade your helm configuration:
+3. Install your helm configuration:
 ```bash
 $ helm install -f values/values.dynamic_config.yaml temporaltest . --timeout 900s
 ```
-
+Note that if you already have a running cluster you could use upgrade command to change dynamic config values:
+```bash
+$ helm upgrade -f values/values.dynamic_config.yaml temporaltest . --timeout 900s
+```
 
 ## Uninstalling
 
--- a/README.md
+++ b/README.md
@@ -296,6 +296,18 @@ Forwarding from [::1]:8081 -> 3000
 
 3. Navigate to the forwarded Grafana port in your browser (http://localhost:8081/), login as `admin` (using the password from step 1), and click on the "Home" button (upper left corner) to see available dashboards.
 
+### Updating dynamic config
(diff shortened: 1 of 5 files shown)


pauldemarco:
Does this include the ability to dynamically add elasticsearch search attribute keys?

vitarb:
@pauldemarco yes, you should be able to add/update `frontend.validSearchAttributes` fields similar to how it's done in the [development.es](https://github.com/temporalio/temporal/blob/master/config/dynamicconfig/development_es.yaml#L7) config.
Then you simply run `helm upgrade ...` and get your cluster config refreshed.

templates/server-configmap.yaml
```diff
@@ -178,4 +178,16 @@ data:
 
     publicClient:
       hostPort: "{{ include "temporal.componentname" (list . "frontend") }}:{{ .Values.server.frontend.service.port }}"
+
+    dynamicConfigClient:
+      filepath: "/etc/temporal/config/dynamic_config.yaml"
+      pollInterval: "10s"
+
+  dynamic_config.yaml: |-
```
markmark206, line 186: I think the general user experience we have been targeting with helm charts is that folks don't need to clone / edit helm chart code, and that they can simply provide the configuration they need via command line parameters, e. g. 

`$ helm install --set server.config.persistence.visibility.sql.user=markmark -f mycustomsettings.yml ...`

This change is a departure from that general approach --to update dynamic configs people need to make a change to chart itself.

This is not the end of the world, just want to call this out, to make sure this is an explicit decision.

markmark206: Actually, I wonder if we can continue to allow people to specify overridable dynamic config values in values.yml and then just "source" their configuration here. Something along the lines of:


`values.yml` change:

```
...
+  dynamic_config:
+            matching.numTaskqueueReadPartitions:
+                    5
   config:
...
```

and in this file, right here:
```
dynamic_config.yaml: |-
    {{- toYaml .Values.server.dynamic_config | nindent 4 }}
```

this way, folks can simply provide dynamic config values via `--set ...` command line option or via `-f mydynamicvalues.yml`, and they don't have to edit helm chart code to change their dynamic configs.

underrun (commented):
my usage with helm in general tends to be to render to yaml rather than run upgrade and i like the suggestion @markmark206 made for specifying things on cli or in a separate file.

it might also be nice if we can avoid duplicating configs in heml/k8s specific way and generate config maps from the same template format used by temporal and/or the docker image itself.

vitarb:
Thanks for the suggestion , I moved config into the separate file and tested that everything works when file is passed from the command line. 
The only caveat was handling empty case properly as we could end up with a file containing "null" string.

For the context here is usage example:
1. Populate dynamic config file into `dc.yaml` containing:
```
server:
  dynamicConfig:
  ....
```
2. Install or upgrade helm passing `-f dc.yaml` option.

I'm quite satisfied with the current result, the last thing we should do is update the documentation.

markmark206:
This looks great, thank you!

My remaining question is, of course, ;) "how do we test this?"

Currently, our pipelines test everything we claim in our readme.

One option I see is to include a basic dynamic values file (e. g. task queue partition configuration):

`values/values.dynamic_configs.yml`:

~~~
matching.numTaskqueueReadPartitions:
- value: 5
  constraints: {} 
~~~

~~~
matching.numTaskqueueWritePartitions:
- value: 5
  constraints: {}
~~~

and use that file (`helm install ... -f values/values.dynamic_configs.yml ...`) in one of our pipelines, such as byoe pipeline (since it is specifically focusing on testing helm chart features):

https://github.com/temporalio/cicd/blob/b2a4160f1223fa02285306192fc7e1c5333ad30f/tasks/deploy-environment-byoe.yml#L65-L79

This would 
* let us make sure that this logic works as expected (or, at least, that `helm install` doesn't immediately fail, even if we don't verify that the file got placed into containers), and 
* will also give our users an easy example to work from.

README.md
```diff
@@ -296,6 +296,15 @@ Forwarding from [::1]:8081 -> 3000
 
 3. Navigate to the forwarded Grafana port in your browser (http://localhost:8081/), login as `admin` (using the password from step 1), and click on the "Home" button (upper left corner) to see available dashboards.
 
+### Updating dynamic config
+By default dynamic config is empty, if you want to override some properties for your cluster, you should:
+1. Create a yaml file with your config (for example dc.yaml).
+2. Populate it with some values under server.dynamicConfig prefix (use the sample provided at `values/values.dynamic_config.yaml` as a starting point)
+3. Install or upgrade your helm configuration:
```
markmark206, line 303: it would be great to be more specific here about when to run `helm install` and when to run `helm upgrade`, and include sample / working command lines for both.

PR #63, opened 2020-08-27
Enable archival features

feedmeapples:


markmark206:
i forgot the context for this -- did we say we wanted to shelve this for now?

PR #65, opened 2020-08-28
Configure cassandra replication factor

markmark206:
Allow for configuring Cassandra replication factor. Example:

```
$ helm install temporaltest --set cassandra.config.cluster_size=5 --set server.config.persistence.default.cassandra.replicationFactor=3 --set server.config.persistence.visibility.cassandra.replicationFactor=3 --set server.replicaCount=3 . --timeout 15m
```

Side note:

With this change, when running the example above, it takes longer for the Cassandra cluster to reach full availability –– now **3** instances of Cassandra need to be online before data could be inserted, as opposed to just one.

During my testing I noticed that the longer Cassandra "time to full availability" started triggering frequent "table {schema_version | schema_update_history} already exist" failures.

The ability to create keyspaces and tables comes online before the ability to create tables, so initial temporal-cassandra-tool runs fail mid-flight, when attempting to insert table, after having created keyspaces and tabls. Subsequent reruns fail because the tables already exist, so db never fully initializes. This change addresses this problem: https://github.com/temporalio/temporal/pull/697.

### DIFF
--- a/templates/server-job.yaml
+++ b/templates/server-job.yaml
@@ -54,7 +54,7 @@ spec:
           image: "{{ $.Values.server.image.repository }}:{{ $.Values.server.image.tag }}"
           imagePullPolicy: {{ $.Values.server.image.pullPolicy }}
           {{- if eq (include "temporal.persistence.driver" (list $ $store)) "cassandra" }}
-          args: ['sh', '-c', 'temporal-cassandra-tool create -k {{ $storeConfig.cassandra.keyspace }}']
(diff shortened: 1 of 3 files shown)


underrun (approved):

mastermanu:
We should create 5 instances of Cassandra (instead of 3) if we are going with 3 replicas

markmark206:
> We should create 5 instances of Cassandra (instead of 3) if we are going with 3 replicas

Yes, the helm chart itself defaults to cassandra cluster size of 3  of cassandra + replication factor of 1.

In the tests that use replication factor of 3, we are deploying Cassandra cluster size of 5 (here is the private repo change: https://github.com/temporalio/cicd/pull/138 , which will take effect when this PR gets merged).

PR #68, opened 2020-08-31
add more antiaffinity rules

underrun:
to prevent temporal services from running on nodes with datastores
and some other services.

### DIFF
--- a/values/values.antiaffinity.yaml
+++ b/values/values.antiaffinity.yaml
@@ -1,49 +1,96 @@
+cassandra:
+  affinity:
+    podAntiAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+      - labelSelector:
+          matchExpressions:
+          - key: app.kubernetes.io/component
+            operator: In
+            values:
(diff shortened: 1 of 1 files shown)


values/values.antiaffinity.yaml
```diff
@@ -1,49 +1,96 @@
+cassandra:
+  affinity:
+    podAntiAffinity:
+      requiredDuringSchedulingIgnoredDuringExecution:
+      - labelSelector:
+          matchExpressions:
+          - key: app.kubernetes.io/component
+            operator: In
+            values:
```
mastermanu, line 9: should we also include history in this as well?

values/values.antiaffinity.yaml
```diff
@@ -1,49 +1,96 @@
...
+            - frontend
```
underrun, line 10: this should maybe include all the temporal pods?

shawnhathaway: Yes, but this functions without via transitive anti affinity for now

PR #73, opened 2020-09-11
Expose SQL Conn Settings via Helm

shawnhathaway:
Am I doing this right?

These should be configurable for consumers of helm as these are exposed via yaml config and in docker config_template

### DIFF
--- a/templates/server-configmap.yaml
+++ b/templates/server-configmap.yaml
@@ -43,6 +43,9 @@ data:
             connectProtocol: "tcp"
             user: {{ include "temporal.persistence.sql.user" (list . "default") }}
             password: {{ `{{ .Env.TEMPORAL_STORE_PASSWORD }}` }}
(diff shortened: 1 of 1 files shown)


markmark206 (approved):
didn't try, but the changes look good, i would recommend running `helm template temporaltest  .` or some such to see that the templates get expanded as you would expect.

sagikazarmark:
This change is not necessary, it should already work if you pass those parameters to the persistence config.

```
{{- with (omit .Values.server.config.persistence.default.sql "driver" "driverName" "host" "port" "connectAddr" "connectProtocol" "database" "databaseName" "user" "password" "existingSecret") }}
{{- toYaml . | nindent 12 }}
{{- end }}
```

The above lines are responsible for filtering all values that we set manually in the template, and add the rest as yaml to the configuration. So if you add `maxConns` to the sql driver config, it should show up in the generated config.

PR #82, opened 2020-10-13
README: accessing Web UI, various text tweaks

markmark206:
Minor README cleanup + adding a section on accessing temporal Web UI.

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,7 +355,7 @@ Forwarding from [::1]:8081 -> 3000
 
 3. Navigate to the forwarded Grafana port in your browser (http://localhost:8081/), login as `admin` (using the password from step 1), and click on the "Home" button (upper left corner) to see available dashboards.
 
-### Updating dynamic config
+### Updating Dynamic Configs
 By default dynamic config is empty, if you want to override some properties for your cluster, you should:
 1. Create a yaml file with your config (for example dc.yaml).
 2. Populate it with some values under server.dynamicConfig prefix (use the sample provided at `values/values.dynamic_config.yaml` as a starting point)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -293,17 +303,31 @@ Bad binaries to reset:
...
-$ kubectl port-forward services/temporaltest-frontend-headless 7233:7233 
+$ kubectl port-forward services/temporaltest-frontend-headless 7233:7233
 Forwarding from 127.0.0.1:7233 -> 7233
 Forwarding from [::1]:7233 -> 7233
 ```
 
-and, from a separate window, use the local port to access the service.
+and, from a separate window, use the local port to access the service from your application or Temporal samples.
+
+### Forwarding Your Machine's Local Port to Temporal Web UI
```
underrun, line 318: good to call this out!

README.md
```diff
@@ -344,6 +368,8 @@ Note that if you already have a running cluster you could use upgrade command to
 $ helm upgrade -f values/values.dynamic_config.yaml temporaltest . --timeout 900s
 ```
 
+Note that doing so will trigger a rolling upgrade of your system.
```
underrun, line 371: i'm not super familiar with helm, but would you want to add the `-f values/values.dynamic_config.yaml` /in addition/ to everything that was specified when initially deployed? otherwise wouldn't it use the defaults and change all sorts of things?

markmark206: Do you mean in all other sections, outside of "Updating Dynamic Config"?

I think the initial statement in this section and point # 3 (which has an example of deploying with dynamic configs) is probably sufficient – it keeps dynamic config complexity out of  other sections, while providing sufficient information to those who are specifically interested in dynamic configs.

What do you think?

cullywakelin (approved):
From my perspective - looks good.

README.md
markmark206: (I will go ahead and merge for velocity, but please let me know if you have more thoughts / feedback on this).

underrun: sgtm

PR #87, opened 2020-10-20
Fix TEMPORAL_GRPC_ENDPOINT value

shkrid:
Service's can reach each other without FQDN
Removed not needed `svc.cluster.local` suffix as it prevents using custom k8s cluster's domains.

### DIFF
--- a/templates/web-deployment.yaml
+++ b/templates/web-deployment.yaml
@@ -35,7 +35,7 @@ spec:
           imagePullPolicy: {{ .Values.web.image.pullPolicy }}
           env:
             - name: TEMPORAL_GRPC_ENDPOINT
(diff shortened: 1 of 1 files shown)


templates/web-deployment.yaml
```diff
@@ -35,7 +35,7 @@ spec:
           imagePullPolicy: {{ .Values.web.image.pullPolicy }}
           env:
             - name: TEMPORAL_GRPC_ENDPOINT
-              value: "{{ include "temporal.fullname" . }}-frontend.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.server.frontend.service.port }}"
```
underrun, line 38: might be better just removing the `cluster.local` as that's the configurable bit

underrun (changes requested):

templates/web-deployment.yaml
shkrid: Fixed.

underrun (approved):
thank you for the patch!

PR #92, opened 2020-10-24
fix metrics relabeling template

underrun:
so that it works and now has default of empty list


underrun:
merging because i need this

PR #95, opened 2020-10-28
Splits Dynamic Config file mount into a separately mounted ConfigMap volume 

mastermanu:
- Splits the singular ConfigMap into two separate ConfigMaps.
- Mounts the dynamic_config.yaml file onto a separate read-only volume that can detect configuration change updates on the fly.
- Changes the default configuration to search for the dynamic_config.yaml file in the newly mounted volume.
- Updates README instructions.

Validated with private deployment

### DIFF
--- a/README.md
+++ b/README.md
@@ -433,7 +433,7 @@ If a rolling upgrade is not desirable, you can also generate the ConfigMap file 
 ```bash
 $ kubectl apply -f dynamicconfigmap.yaml
 ```
-You can use helm upgrade with the "--dry-run" option to generate the template for the dynamicconfigmap.yaml.
+You can use helm upgrade with the "--dry-run" option to generate the content for the dynamicconfigmap.yaml.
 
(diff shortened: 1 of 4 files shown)


underrun (approved):

markmark206:
it'd be super-great to have a test running in the pipeline for this. 

here is how we currently test "dynamic" config updates via rollouts:

passing in dynamic config file: 
https://github.com/temporalio/cicd/blob/68b35a4943e7b3b3944bd7908d16b53244f0b020/tasks/deploy-environment-byoe.yml#L100

verifying that the dynamic config took effect:
https://github.com/temporalio/cicd/blob/051756a04c7a5360355ab6df30179ecf3037c8c9/tasks/run-tests.yml#L93-L98

one way to test this is to append two tasks to the current sequence that runs those tests https://github.com/temporalio/cicd/blob/68b35a4943e7b3b3944bd7908d16b53244f0b020/pipelines/nightly-cassandra-es-byoe.yml#L140-L277

one task would basically perform the update using this config map mechanism (e. g. to 12 tq partitions), and the other tasks would basically be a copy of run-tests task (making sure that the change took effect), but perhaps parametrized by # of expected partitions. i recommend creating a ticket for this work, so we have this functionality covered by tests.

PR #97, opened 2020-10-30
helm chart enhancements

ermeaney:
I just went through a deployment of tee chart and found a few enhancements like port naming for istio protocol auto discovery and the need to add some volumes

Also I was having problems with the ingress deploy which seemed better after changing the formatting a bit


underrun:
This looks good, thanks for the contribution!

PR #98, opened 2020-10-31
Update README.md with guidance on setting up database with k8s pod

MrSaints:
The documentation suggests setting up the database for Temporal
using `temporal-sql-tool` from another repository.

But it is also possible to set it up from within a pod in Kubernetes
using the Temporal server image (as the binary is copied into it).

This may be better as it requires less set-up, and the Kubernetes
cluster likely has easier access to the target database (networking-wise).


underrun:
this has been moved out of the server image but should be in the admin tools image. at the same time we're changing he way we approach helm charts so the way something like this is run might be different depending on how / if we deploy admin tools by default. See https://github.com/temporalio/proposals/issues/42 for more details on pending helm chart changes.

PR #99, opened 2020-10-31
fix issue with additionaVolumes / Mounts

underrun:
they are broken if missing - omit if none specified


underrun:
merged to unblock pipelines - note that we could add a test pipeline to this to run `helm template` to catch these sorts of issues.

PR #101, opened 2020-11-02
Adds ability to set Elastic Search connection parameters via helm chart

mastermanu:


templates/server-deployment.yaml
```diff
@@ -72,11 +72,17 @@ spec:
...
+              value: "{{ $.Values.elasticsearch.host }}"
             - name: ES_PORT
               value: "{{ $.Values.elasticsearch.port }}"
             - name: ES_SCHEME
               value: "{{ $.Values.elasticsearch.scheme }}"
+            - name: ES_VIS_INDEX
+              value: "{{ $.Values.elasticsearch.visibilityIndex }}"
+            - name: ES_USER
+            - value: "{{ $.Values.elasticsearch.username }}"
+            - name: ES_PWD
```
mastermanu, line 84: should be able to re-use the temporal-visibility-store-password here below

templates/server-configmap.yaml
```diff
@@ -77,8 +77,10 @@ data:
                 url:
                     scheme: "{{ $.Values.elasticsearch.scheme }}"
                     host: "{{ $.Values.elasticsearch.host }}:{{ $.Values.elasticsearch.port }}"
+                username: "{{ $.Values.elasticsearch.username }}"
```
underrun, line 80: does this still work if the ES endpoint doesn't require auth?

templates/server-configmap.yaml
```diff
@@ -77,8 +77,10 @@ data:
...
+                password: "{{ $.Values.elasticsearch.password }}"
```
mastermanu, line 81: should ideally pull this from environment variable

templates/server-configmap.yaml
mastermanu: Good question. Yes, because the ES client library checks for empty string on username/password

markmark206 (approved):

PR #103, opened 2020-11-03
HPA support for the various services

ermeaney:


templates/server-configmap.yaml
```diff
@@ -77,10 +77,8 @@ data:
                 url:
                     scheme: "{{ $.Values.elasticsearch.scheme }}"
                     host: "{{ $.Values.elasticsearch.host }}:{{ $.Values.elasticsearch.port }}"
-                username: "{{ $.Values.elasticsearch.username }}"
-                password: "{{ $.Values.elasticsearch.password }}"
                 indices:
-                    visibility: "{{ $.Values.elasticsearch.visibilityIndex }}"
+                    visibility: temporal-visibility-dev
```
chitreshd, line 81: why are we hardcoding?

values.yaml
```diff
@@ -156,6 +162,12 @@ server:
 
   history:
     # replicaCount: 1
+    autoscaling:
+      enabled: false
+      minReplicas: 1
+      maxReplicas: 3
```
chitreshd, line 168: for prod you would need to start with atleast 3 replicas

values.yaml
```diff
@@ -174,6 +186,12 @@ server:
 
   matching:
     # replicaCount: 1
+    autoscaling:
+      enabled: false
+      minReplicas: 1
```
chitreshd, line 191: same as history

values.yaml
```diff
@@ -287,9 +311,6 @@ elasticsearch:
   host: elasticsearch-master-headless
   scheme: http
   port: 9200
-  username: ""
-  password: ""
-  visibilityIndex: "temporal-visibility-dev"
 
 prometheus:
   enabled: true
```
chitreshd, line 316: i thought we wanted to disable prometheus

underrun:
Thanks for the contribution! Please remember to sign the CLA.

For this specific PR, we've recently landed on an alternate approach to helm charts: https://github.com/temporalio/proposals/issues/42 - as a result, and because HPA resources can be added independently of temporal itself, we will not be merging this at this time.

PR #107, opened 2020-11-07
support sidecar containers and provide cloudsqlproxy example

mnichols:
This PR extends the helm templates to support sidecar containers. An example for adding a Google cloud sql proxy sidecar is provided.

### DIFF
--- a/values/values.cloudsqlproxy.yaml
+++ b/values/values.cloudsqlproxy.yaml
@@ -1,0 +1,20 @@
+server:
+  sidecarContainers:
+    - name: cloud-sql-proxy
+      image: gcr.io/cloudsql-docker/gce-proxy:1.17
+      command:
+        - "/cloud_sql_proxy"
+        - "-ip_address_types=PRIVATE"
+        - "-instances=_PROJECTNAME_:_REGION_:_INSTANCENAME_=tcp:5432"
+        - "-credential_file=/etc/google-cloud-key/key.json"
+      securityContext:
+        runAsNonRoot: true
+      volumeMounts:
+        - name: google-cloud-key
+          mountPath: /etc/google-cloud-key
+          readOnly: true
+
+  additionalVolumes:
+    - name: google-cloud-key
+      secret:
+        secretName: cloud-sql-proxy-sa
--- a/README.md
+++ b/README.md
@@ -96,6 +96,18 @@ Other components (Prometheus, Kafka, Grafana) can be omitted from the installati
     temporaltest . --timeout 900s
 ```
 
+### Install with sidecar containers
+
+You may need to provide your own sidecar containers. 
+
+To do so, you may look at the example for Google's `cloud sql proxy` in the `values/values.cloudsqlproxy.yaml` and pass that file to `helm install`. 
+
+Example:
+
+```bash
+~/temporal-helm$ helm install -f values/values.cloudsqlproxy.yaml temporaltest . --timeout 900s
+```
+
 ### Install with your own ElasticSearch
 
 You might already be operating an instance of ElasticSearch that you want to use with Temporal.
--- a/templates/server-deployment.yaml
+++ b/templates/server-deployment.yaml
@@ -126,6 +126,10 @@ spec:
             {{- end }}
           resources:
             {{- toYaml (default $.Values.server.resources $serviceValues.resources) | nindent 12 }}
+        {{- if $.Values.server.sidecarContainers }}
+        {{- toYaml $.Values.server.sidecarContainers | nindent 8 }}
+        {{- end }}
+
       {{- with $.Values.imagePullSecrets }}
       imagePullSecrets:
       {{- toYaml . | nindent 8 }}
--- a/values.yaml
+++ b/values.yaml
@@ -7,6 +7,7 @@ debug: false
 
 server:
   enabled: true
+  sidecarContainers:
   image:
     repository: temporalio/server
     tag: 1.8.0


underrun:
Thanks for the PR!

My initial thoughts are that it would be really nice if users could apply values files without editing. Maybe just an example in the docs would be useful?

mnichols:
Hi @underrun! Helm isn't really my wheelhouse and I wasn't sure of a way to apply a sidecar to the existing templates without the approach taken here. 
How does one extend the deployment without doing the `-f <valuesfile>` approach  used so far in this repo?
Or were you just saying to _not_ include the `cloudsqlproxy` values file and make it purely a form of documentation? 

Personally I don't mind the approach in the `values.postgres.yaml` and others but could see how they should be maintained outside. Only question I'd have is how best to keep it maintained in sync with the helm templates they are extending.

underrun:
I meant the documentation angle... But let me think about this a bit and chat with some folks next week.

Thanks again for working on this!

mnichols:
Sure thing..lemme know what y'all need. I was just pleased to be able to simply support cloudsqlproxy for our prod stuff in GCP. Your work here made it a snap to extend...nice job!

jpearll:
Hi,
I am new with Helm. I am trying to deploy temporal on GCP (GKE).

I have tried to use the PR for getting Cloud Proxy support, with the following command:
`helm install -f ./values/values.cloudsqlproxy.yaml workflow . --timeout 15m`

But I have systematically a status 'Does not have minimum availability' on GCP for the temporal-frontend, temporal-worker, temporal-history, temporal-matching and temporal-worker workloads. Even if I tried to add nodes (6 nodes, auto-scaling).

I am not sure about how to tackle this. Also, what is the recommended GKE configuration for the cluster/node pool/machine types?

Thanks !

mnichols:
> Hi,
> I am new with Helm. I am trying to deploy temporal on GCP (GKE).
> 
> I have tried to use the PR for getting Cloud Proxy support, with the following command:
> `helm install -f ./values/values.cloudsqlproxy.yaml workflow . --timeout 15m`
> 
> But I have systematically a status 'Does not have minimum availability' on GCP for the temporal-frontend, temporal-worker, temporal-history, temporal-matching and temporal-worker workloads. Even if I tried to add nodes (6 nodes, auto-scaling).
> 
> I am not sure about how to tackle this. Also, what is the recommended GKE configuration for the cluster/node pool/machine types?
> 
> Thanks !

@jpearll I cant comment on the recommended GKE config you need, but it sounds like if your pods arent starting up with this cloudsqlproxy config in place your attempt to connect to your db might be hanging. This is a sidecar and I've found they tend to fail badly (eg silently). You might want to check a direct db connection to your cloudsql instance first before trying with this proxy in place?

values/values.cloudsqlproxy.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+        # If connecting from a VPC-native GKE cluster, you can use the
+        # following flag to have the proxy connect over private IP
+        - "-ip_address_types=PRIVATE"
+
+        # Replace DB_PORT with the port the proxy should listen on
+        # Defaults: MySQL: 3306, Postgres: 5432, SQLServer: 1433
+
+        - "-instances=<YOUR_PROJECT>:us-central1:<YOUR-INSTANCE-NAME>=tcp:5432"
+      # [START cloud_sql_proxy_k8s_volume_mount]
```
underrun, line 18: this looks like it's maybe copied from https://github.com/GoogleCloudPlatform/cloudsql-proxy/blob/master/examples/kubernetes/proxy_with_sa_key.yaml or something else in that repo - maybe we can match the style of the other values files here without comments etc? otherwise we'd need an apache 2 notice header and that seems a little distracting for an example file.

underrun (changes requested):
My apologies for the much-longer-than-a-week it took me to come back to this. We'd like to try and merge this in with a little cleanup. Thanks again for putting this together!

jasonzoladz:
@underrun, what's the ETA for merging this?  We need this as well.  :)

underrun:
@mnichols would you like to make the requested change? If not I can open a parallel PR to get this merged and close this one? Either way is totally fine. Thanks again for your work on this!

values/values.cloudsqlproxy.yaml
mnichols: ok now i'm the slow one. i'll remove the comment and match the style

mnichols:
@underrun I made the changes I think you were after and squashed it after rebasing on latest master

JDWardle:
@underrun Is there anything blocking this from being merged in? This would be very helpful to have not just for cloudsql but support for adding in any other sidecars as well.

hazcod:
@mnichols @underrun any news here?

mnichols:
🤷 
@underrun ?

underrun (approved):

ThomaciousD:
🚨 FYI: The `server.sidecarContainers` pattern is no longer supported in the chart. Sidecar support was removed in https://github.com/temporalio/helm-charts/pull/649. With current versions, you need to use `server.additionalInitContainers` and set `restartPolicy: Always` for the proxy to behave like a sidecar (Kubernetes restartable init sidecars). 

The old example:

```
  server:
    sidecarContainers:
      - name: cloud-sql-proxy
        image: gcr.io/cloudsql-docker/gce-proxy:1.17
        command:
          - "/cloud_sql_proxy"
          - "-ip_address_types=PRIVATE"


```  
**will be ignored by the chart**; the proxy won’t show up in the server pods. Switching to additionalInitContainers with `restartPolicy: Always` is the supported approach now.

PR #112, opened 2020-11-17
Fix yaml parsing error

vitarb:
Fixes YAML indentation that was broken since #92
```
Error: YAML parse error on temporal/templates/server-service-monitor.yaml: error converting YAML to JSON: yaml: line 19: could not find expected ':'
```


markmark206 (approved):
i am very puzzled as to how this works in our pipelines but not on our laptops, but the change looks good.

PR #117, opened 2020-12-08
Allow custom annotations

### LINKED ISSUE #116: annotations values / loadbalancer type are missing for the services
I need to access Temporal from outside the cluster for the web with the ingress and for the rpc service
We have a metalLB with external-DNS setup, but for Temporal, some customizations are missing:

- custom annotations for the services
- service type load balancer

lerminou:
This change allow to set custom annotations for the frontend service, in my case add an external-dns annotation
The helper is a copied from bitnami common charts so any annotations can be refactored with this method (and add more if they are missing)
fix #116

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -311,3 +311,17 @@ The first Cassandra host in the stateful set.
 {{- printf "%s-kafka-headless:9092" .Release.Name -}}
 {{- end -}}
 {{- end -}}
+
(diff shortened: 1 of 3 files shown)


lerminou:
Hi, 
@wxing1292 @mastermanu  @underrun 
Is there a change to check this PR please ?

wxing1292 (approved):

PR #120, opened 2020-12-17
Replace deprecated stable chart repo with current ones

deltasquare4:
* The dependent charts are no longer being served from
  old stable repo. Update them to the current functional
  URLs.
* Versions are unchanged for compatibility


underrun (approved):

underrun:
thanks!

PR #130, opened 2021-01-18
add service.serviceMonitor options

xxbbxb:
* `nindent 6` worked for me, because yaml renders in next line
* `namespaceOverride` for cases when prom-operator installed in another namespace


AmetDevops:
Thanks!
Works like a charm

underrun:
this has been fixed in another PR

PR #136, opened 2021-02-03
add service accounts to deployments

jfrey:
This change provides a service account per deployment to allow the use of Consul Connect Inject with ACL's enabled.

While this will enable the Consul service mesh to work it also is probably appropriate to allow configuration of the service bind address to configure proxy only access to the services.  That change is more invasive as there appears to be a gossip port for membership as well as an rpc port.

Any advice on how you'd like that addressed potentially would be good.


underrun:
Thanks for the contribution - the way our helm charts will be handled going forward will be scoped down to just temporal services. service accounts will be possible to support but in a different way. See this for more details: https://github.com/temporalio/proposals/issues/42

PR #138, opened 2021-02-09
Fix indentation of annotations metadata for the `web` templates

jverce:
Changed the `web-ingress` and `web-service` templates to adapt the
annotations metadata subsection to match what the [Helm
docs](https://github.com/helm/charts/blob/master/REVIEW_GUIDELINES.md#ingress) do.

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -53,7 +53,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.2.0
+version: 0.2.1
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
--- a/templates/web-ingress.yaml
+++ b/templates/web-ingress.yaml
@@ -15,10 +15,10 @@ metadata:
     app.kubernetes.io/version: {{ .Chart.AppVersion | replace "+" "_" }}
     app.kubernetes.io/component: web
     app.kubernetes.io/part-of: {{ .Chart.Name }}
-  {{- with .Values.web.ingress.annotations }}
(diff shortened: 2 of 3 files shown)


Chart.yaml
```diff
@@ -53,7 +53,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.2.0
+version: 0.2.1
```
underrun, line 56: we haven't been bumping version for this and this change doesn't affect output of the chart - just cleanup of the template.

```suggestion
version: 0.2.0
```

underrun (changes requested):
Thanks for the PR - will merge after changes addressed.

Chart.yaml
jverce: This change in fact fixes rendering issues.

To reproduce:

1. Navigate to the root of the project
1. Create a `values-web.yaml` with the following content:
```yaml
web:
  service:
    annotations: 
      service.beta.kubernetes.io/aws-load-balancer-backend-protocol: http
      service.beta.kubernetes.io/aws-load-balancer-ssl-negotiation-policy: 'ELBSecurityPolicy-TLS-1-2-2017-01'
```

2. Run the following command:
```bash
$ helm template temporal . -f values.yaml
```

3. The following error is displayed:
```
Error: YAML parse error on temporal/templates/web-service.yaml: error converting YAML to JSON: yaml: line 14: did not find expected key

Use --debug flag to render out invalid YAML
```

After the change in this PR, the template is generated successfully.

underrun: awesome thanks!

underrun (approved):

PR #141, opened 2021-02-17
Update to 1.7.0

Ardagan:


Chart.yaml
```diff
@@ -53,8 +53,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.2.1
```
alexshtin, line 56: I don't remember I have ever changed this.

alexshtin (approved):

Chart.yaml
Ardagan: Comment above this number states that it should be changed whenever app version changes.

PR #150, opened 2021-03-24
Release 1.8.0

Ardagan:
## Description, Motivation and Context

> Thanks for opening a PR to our repos! If it is a significant code change, please make sure there is an
> opened issue where we have accepted the idea before filing this PR.

What was changed:
Updated to reference temporal binaries v1.8.0

Motivation & Context:
Release v1.8.0

Closes issue: 
N/A

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -53,8 +53,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.2.2
+version: 0.2.3
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 1.7.1
+appVersion: 1.8.0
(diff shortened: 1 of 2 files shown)


Chart.yaml
```diff
@@ -53,8 +53,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.2.2
+version: 0.2.3
```
alexshtin, line 56: Shouldn't we sync this version with `appVersion`?

alexshtin (approved):

Chart.yaml
Ardagan: I don't think these have to match. But yes, it should be bumped according to the comment above.

PR #162, opened 2021-04-22
feat: handle mTLS support

lerminou:
## What was changed:

This PR add support to configure mTLS temporal
## Checklist

1. Closes issue: 

2. How was this tested:

Deploy with and without all new options to a RKE cluster with success

3. Any docs updates needed?
Done in the Readme

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -66,6 +66,10 @@ server:
   affinity: {}
   additionalVolumes: []
   additionalVolumeMounts: []
+  additionalEnvVars: []
+  additionalToolsEnvVars: []
+  additionalWebEnvVars: []
+  tls: []
 
   config:
     logLevel: "debug,info"
--- a/values.yaml
+++ b/values.yaml
@@ -73,7 +73,37 @@ server:
   config:
     logLevel: "debug,info"
 
-    tls: []
+    tls: 
+      enabled: false
+      server:
+        certFile: ""
+        keyFile: ""
+        requireClientAuth: true
+        clientCaFiles: []
+      client:
+        serverName: ""
+        rootCaFiles: []
+      internode:
+        enabled: true
+        server:
+          certFile: ""
+          keyFile: ""
+          requireClientAuth: true
+          clientCaFiles: []
+        client:
+          serverName: ""
+          rootCaFiles: []
+      frontend:
+        enabled: true
+        server:
+          certFile: ""
+          keyFile: ""
+          requireClientAuth: true
+          clientCaFiles: []
+        client:
+          serverName: ""
+          rootCaFiles: []
+
 
     # IMPORTANT: This value cannot be changed, once it's set.
     numHistoryShards: 512
--- a/values.yaml
+++ b/values.yaml
@@ -67,12 +67,13 @@ server:
   additionalVolumes: []
   additionalVolumeMounts: []
   additionalEnvVars: []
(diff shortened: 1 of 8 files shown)


values/values.mtls.yaml
```diff
@@ -0,0 +1,121 @@
+server:
+  additionalToolEnvVars:
```
underrun, line 2: Maybe `additionalTctlEnvVars`?

values.yaml
```diff
@@ -66,6 +66,10 @@ server:
   affinity: {}
   additionalVolumes: []
   additionalVolumeMounts: []
+  additionalEnvVars: []
+  additionalToolsEnvVars: []
+  additionalWebEnvVars: []
+  tls: []
```
underrun, line 72: since this drops into the configmap - maybe move `tls` it to the config section?

values/values.mtls.yaml
```diff
@@ -0,0 +1,121 @@
...
+      subPath: temporal-tools.key
+    - name: cert
+      mountPath: /etc/temporal/certs/temporal-web.crt
+      subPath: temporal-web.crt
+    - name: key
+      mountPath: /etc/temporal/certs/temporal-web.key
+      subPath: temporal-web.key
+
+  tls:
+    tls:
```
underrun, line 99: this double tls is a little confusing to me. can we rework the configmap - it looks like there's a hard coded `tls:` in the configmap already so do we need it here?

also since this is entirely in the server config, maybe it'd be better to put it at `server.config.tls` instead of `server.tls`?

underrun (commented):
Thanks for the contribution!

One overall question on the approach: would it be more ergonomic to either structure the tls template in the config map a little more and add the nested internode and frontend sections (which users could specify either one or both of) or make a more generic splat for additionalConfig or something like that?

the structured approach would be more helm idiomatic and provide better checking for things, but the splat would allow looser coupling of the helm chart and the server config which could be useful too ... thoughts?

values/values.mtls.yaml
lerminou: yeah actually it's already done in the configMap to set the global  tls key , I forgot to remove this one.

values.yaml
lerminou: ok i will

lerminou:
Thanks for the review, comments addressed.

For the global approach, I can work for a more atomic configuration.

lerminou:
@underrun , I reworked the configMap to be more atomic, WDYT ?
First option is to configure TLS globally, and we can choose every parameter for both internode & frontend

it's my first Helm project, don't hesitate if I use wrong practices.

lerminou:
@underrun , do you have time to check this ?

underrun:
Based on our new approach to structuring our helm chart this will be supported in a different way. See https://github.com/temporalio/proposals/issues/42 for more details on what's changing.

gthomson31:
I had went with customisation of the existing chart based from the pull request above

joshbranham:
Were you able to get the env vars for configuring TLS working? In my testing with the latest helm chart, the `config_template.yaml` does not seem to include the optional env var defaults as shown [here](https://github.com/temporalio/temporal/blob/main/docker/config_template.yaml)

Template in helm: https://github.com/temporalio/helm-charts/blob/master/templates/server-configmap.yaml

PR #164, opened 2021-04-23
Configurable archival and namespaceDefaults settings.

CGreenburg:
## What was changed:

- Added the ability to set archival and namespaceDefaults in values file. If not provided, archival is disabled (like before) and namespaceDefaults isn't set.

### DIFF
--- a/templates/server-configmap.yaml
+++ b/templates/server-configmap.yaml
@@ -186,8 +186,20 @@ data:
       policy: "noop"
       toDC: ""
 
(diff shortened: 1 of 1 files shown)


underrun:
Thanks for the contribution!

Due to the recent decision to change our approach to helm charts we will not be merging this as there will be a different method for supplying all relevant temporal configuration. See this for more details: https://github.com/temporalio/proposals/issues/42

PR #168, opened 2021-04-26
Allow loadbalancerIP for frontend service

lerminou:
## Why?

It allows static IP from a load balancer 

## Checklist

2. How was this tested:

deploy the helm with the values set, OK

### DIFF
--- a/templates/server-service.yaml
+++ b/templates/server-service.yaml
@@ -15,8 +15,8 @@ metadata:
   annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.server.frontend.service.annotations "context" $) | nindent 4 }}
   {{- end }}
 spec:
-  {{- with .Values.server.frontend.service.loadBalancerIP }}
-  loadBalancerIP: {{.}}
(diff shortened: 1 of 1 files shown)


nkranes-frameio (approved):

templates/server-service.yaml
```diff
@@ -15,6 +15,9 @@ metadata:
   annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.server.frontend.service.annotations "context" $) | nindent 4 }}
   {{- end }}
 spec:
+  {{- if .Values.server.frontend.service.loadBalancerIP }}
+  loadBalancerIP: {{ .Values.server.frontend.service.loadBalancerIP }}
```
underrun, line 19: this only works if `.Values.server.frontend.service.type` is `LoadBalancer` - can we check both things for safety?

underrun:
documentation in readme or comment in values file would be helpful for usage here

underrun:
closing this after conversations about how we will handle these resources going forward. see: https://github.com/temporalio/proposals/issues/42 for more details.

Issue #182, opened 2021-05-20
[Bug] service init containers not finishing due to waiting for default keyspace to become ready


mickmcgrath13:
**Describe the bug**
Installing the helm chart as follows:
- elasticsearch disabled
- mysql disabled
- prometheus disabled
- grafana enabled
- cassandra enabled
- cassandra persistence disabled

Got an error in many of the services' (frontend, history, matching, etc) init containers when installing:
```
waiting for default keyspace to become ready
```

Apparently, the keyspace becomes ready with the Job: https://github.com/temporalio/helm-charts/blob/master/templates/server-job.yaml

The job has the following annotations:
```
  annotations:
    {{- if .Values.cassandra.enabled }}
    "helm.sh/hook": post-install
    {{- else }}
    "helm.sh/hook": pre-install
    {{- end }}
```

..it seems that the post-install hook doesn't execute until after the pods are ready, which they don't because they're waiting for this job to run.

The job already has init containers that are waiting for cassandra to come up, so I'm not sure the install hooks are necessary.

**To Reproduce**
1. create a temporal namespace
2. use the following values.yaml
```
server:
  enabled: true
  replicaCount: 1
  config:
    persistence:
      default:
        driver: "cassandra"
        cassandra:
          hosts: ["temporal-cassandra.temporal.svc.cluster.local"]
          # port: 9042
          keyspace: "temporal"
          user: "user"
          password: "password"
          existingSecret: ""
          replicationFactor: 1
          consistency:
            default:
              consistency: "local_quorum"
              serialConsistency: "local_serial"
      visibility:
        driver: "cassandra"

        cassandra:
          hosts: ["temporal-cassandra.temporal.svc.cluster.local"]
          keyspace: "temporal_visibility"
          user: "user"
          password: "password"
          existingSecret: ""
          replicationFactor: 1
          consistency:
            default:
              consistency: "local_quorum"
              serialConsistency: "local_serial"
  frontend:
    replicaCount: 1

  history:
    replicaCount: 1

  matching:
    replicaCount: 1

  worker:
    replicaCount: 1

admintools:
  enabled: true
web:
  enabled: true
  replicaCount: 1
schema:
  setup:
    enabled: true
    backoffLimit: 100
  update:
    enabled: true
    backoffLimit: 100
elasticsearch:
  enabled: false
prometheus:
  enabled: false
grafana:
  enabled: false
cassandra:
  enabled: true
  persistence:
    enabled: false
  config:
[...]

PR #183, opened 2021-05-20
Temporal working with cassandra

mickmcgrath13:
## What was changed
- removed helm hook annotations from the jobs
- add imagePullSecrets to the web deployment
- bump chart version

## Why?
https://github.com/temporalio/helm-charts/issues/182

## Checklist
n/a

1. Closes issue: 
https://github.com/temporalio/helm-charts/issues/182

2. How was this tested:
deployed successfully to k8s cluster

3. Any docs updates needed?
no

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -49,7 +49,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.9.4
+version: 0.9.5
 
 # This is the version number of the application being deployed. This version number should be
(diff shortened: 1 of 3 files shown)


underrun:
thank you for your contribution!

if either of these are still an issue it'd be best to split the PR and address one issue in each.

also please don't forget to sign our CLA!

underrun:
based on conversations about how we're handling helm charts going forward, we will take a different approach to solving these issues if they still exist. see https://github.com/temporalio/proposals/issues/42 for more details.

PR #186, opened 2021-05-23
Update readme to remove aws cli prereq

underrun:
## What was changed

* Update readme to remove aws cli prereq
* Remove wait from install as it is not needed

## Why?

To clean up the chart: aws cli and wait are not needed.


tsurdilo:
@underrun 
is the "--timeout 15m" also needed in the 
Minimal installation section ? https://github.com/temporalio/helm-charts/blob/e2cb20a8875161a5b3c79ef4258398135c6b84c0/README.md#minimal-installation-with-required-dependencies-only

underrun:
@tsurdilo re: the 15 minute timeout - the deployment should happen faster than that - the timeout is really to let you know something failed. Maybe also to set reasonable expectations around how long is too long.

But it's also not strictly necessary - think we should remove it?

underrun:
This will change

PR #193, opened 2021-06-04
Configure opentelemetry metrics and scraper

Ardagan:
## What was changed

Added opentelemetry metrics and prometheus scraper.

## Why?

test opentelemetry metrics

## Checklist

1. Closes 
N/A

2. How was this tested:

N/A

3. Any docs updates needed?

N/A

### DIFF
--- a/templates/server-configmap.yaml
+++ b/templates/server-configmap.yaml
@@ -105,8 +105,13 @@ data:
           tags:
             type: frontend
           prometheus:
+            framework: "opentelemetry"
             timerType: {{ default .Values.server.metrics.prometheus.timerType .Values.server.frontend.metrics.prometheus.timerType }}
             listenAddress: "0.0.0.0:9090"
+          prometheusSDK:
+            framework: "tally"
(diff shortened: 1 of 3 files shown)


templates/server-configmap.yaml
```diff
@@ -105,8 +105,13 @@ data:
           tags:
             type: frontend
           prometheus:
+            framework: "opentelemetry"
             timerType: {{ default .Values.server.metrics.prometheus.timerType .Values.server.frontend.metrics.prometheus.timerType }}
             listenAddress: "0.0.0.0:9090"
+          prometheusSDK:
```
underrun, line 111: what does promtheus SDK mean? - like why is it called that. and why do we need both? do we need to scrape both 9090 and 9091?

Ardagan: We need both. We have to report metrics for SDK via tally due to sdk not using opentelemetry yet. OT and SDK can't serve from same address unfortunately.

underrun (approved):

Issue #200, opened 2021-06-18
[Bug] Unable to disable elastic authentication


eufebius:
**Describe the bug**
If external elasticsearch is used, u cannot disable user authentification and becauser of that init containers will fail during ES checks ( e.g. aws es service )

**To Reproduce**
1. Deploy with external ES without username, password
2. Check check-elasticsearch-index init container in server deployment
3. See error

**Expected behavior**
Add possibility to continue without username, password

robholland:
I believe https://github.com/temporalio/helm-charts/pull/522 fixes this as while username and password will still be sent, they will be sent empty. It is my understanding that Elasticsearch will be content with this if auth was not enabled. Please re-open if this is not the case.

PR #201, opened 2021-06-18
Add option to disable elastic authentication

eufebius:
## What was changed
Add condition to init container which need to check elasticsearch 

## Why?
To enable work with ES without authentication ( aws es service )

## Checklist

1. Closes https://github.com/temporalio/helm-charts/issues/200

2. How was this tested:
- tried deploy with $.Values.elasticsearch.authentication.enabled.true
- tried deploy with $.Values.elasticsearch.authentication.enabled.false

3. Any docs updates needed?

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -291,6 +291,8 @@ elasticsearch:
   port: 9200
   version: "v7"
   logLevel: "error"
+  authentication:
+    enabled: true
   username: ""
   password: ""
   visibilityIndex: "temporal_visibility_v1_dev"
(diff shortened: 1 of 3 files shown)


Whitespace:
@eufebius I think your commit is a different email than the one that your GH account uses, and you have to sign the CLA with the other email.

values.yaml
```diff
@@ -291,6 +291,8 @@ elasticsearch:
   port: 9200
   version: "v7"
   logLevel: "error"
+  authentication:
+    enabled: true
```
underrun, line 295: flip option to `.authentication.disabled` instead - should be enabled by default and if we want to disable this set to true

underrun:
Depending on the final scope of changes that fall under https://github.com/temporalio/proposals/issues/42 this may be handled differently in the future. Ideally the elastic search auth method shouldn't be something that needs to be known by the helm chart, but I'm leaving this open for now so we're tracking that if there are any connections made to ES from inline code in the helm charts we need to account for auth strategy.

underrun:
After talking about this, i don't think we'll need to have connections info directly in kubernetes resources like this going forward. Our new approach will ensure that we can connect to es either with or without user/pass

PR #212, opened 2021-07-20
v1.11.0 release

yiminc:
v1.11.0 release

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -9,7 +9,7 @@ server:
   enabled: true
   image:
     repository: temporalio/server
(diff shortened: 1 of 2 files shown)


values.yaml
```diff
@@ -229,7 +229,7 @@ web:
 
   image:
     repository: temporalio/web
-    tag: 1.10.0
+    tag: 1.11.0
```
wxing1292, line 232: is web also updated?

wxing1292 (approved):

PR #227, opened 2021-09-22
Update metrics config to global

Ardagan:
## What was changed

Per-service metrics config is deprecated. Use global config section instead.

## Checklist

1. Closes 

2. How was this tested:

manual `helm template` run

3. Any docs updates needed?

### DIFF
--- a/templates/server-configmap.yaml
+++ b/templates/server-configmap.yaml
@@ -1,20 +1,21 @@
 {{- if $.Values.server.enabled }}
+{{- range $service := (list "frontend" "history" "matching" "worker") }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: "{{ include "temporal.fullname" . }}-config"
+  name: "{{ include "temporal.componentname" (list $ $service) }}-config"
   labels:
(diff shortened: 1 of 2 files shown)


templates/server-configmap.yaml
```diff
@@ -1,36 +1,37 @@
 {{- if $.Values.server.enabled }}
+{{- range $service := (list "frontend" "history" "matching" "worker") }}
```
underrun, line 2: is the intention here is to create a different config map with a different `config_template.yaml` for every service?

jbreiding: not a different tag but a different tag value?

but that tag value also maps to the env var that directs the service of which service profile to run.

this feels like something that should also be driven from that value.

https://github.com/temporalio/helm-charts/blob/4c7ee04165dc2bfe6dcbb67abbda6c0ac98521e7/templates/server-deployment.yaml#L98

jbreiding (changes requested):
Seems like the variation of tag value should default coming from the service itself.

Ardagan:
I'll close this pr for now. Will update after some more discussion.

PR #242, opened 2021-11-11
Use `emptyDir` for Temporal configuration mount path

### LINKED ISSUE #225: [Bug] Deployment Using Helm Chart Getting Failed
### Describe the bug

Trying to use latest helm chart from this repo, but multiple pods are crashloopbackoff with error logs 
`2021/09/07 03:24:00 unable to create open ./config/docker.yaml: permission denied`
`
image

### To Reproduce

1.  git clone https://github.com/temporalio/helm-charts.git
2. helm dependencies update
3. helm install \
    --set server.replicaCount=1 \
    --set cassandra.config.cluster_size=1 \
    --set prometheus.enabled=false \
    --set grafana.enabled=false \
    --set elasticsearch.enabled=false \
    temporaltest . --timeout 15m

### Versions

- OS: [e.g. Linux]
- Temporal Version [chart version temporal-0.12.0]
- Are you using Docker or Kubernetes or building Temporal from source? 
       Kuberntes via kind and on GKE as well

srikanthchelluri:
## What was changed

This change codifies the suggestion in the following issue/comment:
https://github.com/temporalio/helm-charts/issues/225#issuecomment-932263282

Using an empty directory here is a chart/app version agnostic fix.

## Why?

Details in https://github.com/temporalio/helm-charts/issues/225.

## Checklist

1. Closes #225 

2. How was this tested: we're using a custom version of the Helm chart with this change and it behaves as expected.

3. Any docs updates needed? Unlikely, this should be expected behavior.

### DIFF
--- a/templates/server-deployment.yaml
+++ b/templates/server-deployment.yaml
@@ -127,6 +127,8 @@ spec:
                port: rpc
           {{- end }}
           volumeMounts:
+            - name: config-store
+              mountPath: /etc/temporal/config
             - name: config
               mountPath: /etc/temporal/config/config_template.yaml
               subPath: config_template.yaml
@@ -146,6 +148,8 @@ spec:
       {{- toYaml . | nindent 8 }}
       {{- end }}
       volumes:
+        - name: config-store
+          emptyDir: {}
         - name: config
           configMap:
             name: "{{ include "temporal.fullname" $ }}-config"


sw-yx:
appreciate the contrib and appears that @krousey solved it upstream. pls reopen if mistaken.

PR #243, opened 2021-11-23
Fix wrong metricRelabeling key in Servicemonitor

containerpope:
Helm chart was unable to render due to misconfiguration in service monitor value selector for relabelings


underrun (approved):

underrun:
thanks for the bug fix!

PR #244, opened 2021-11-24
feat: add job to create namespaces post install

oss92:
## What was changed
Add a kubernetes job to create namespaces post-install

## Why?
Makes environments easier to setup and deployments replicable
 
## Checklist

1. Closes 

2. How was this tested:
Used it to create replicable deployment in my organization

3. Any docs updates needed?

### DIFF
--- a/templates/server-job.yaml
+++ b/templates/server-job.yaml
@@ -210,6 +210,47 @@ spec:
       {{- toYaml . | nindent 8 }}
       {{- end }}
 ---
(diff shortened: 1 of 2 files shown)


sw-yx:
appreciate this PR, but unfortunately is off topic for the goals of this repo. However, we're trying to preserve the modifications people make for others to find, so have added an `Alternative` label that we'll be promoting more in coming months as we curate good modifications.

PR #246, opened 2021-11-27
Support adding custom service account

### LINKED ISSUE #245: [Feature Request] One should be able to use custom service account
User should be able to create custom service accounts for temporal pods.

Custom service accounts would be useful in below situations:
- In cluster RBAC using Role and ClusterRole
- Associating public cloud IAM roles with kubernetes service account so that the pods can access cloud services
   - Specifically this would be beneficial when archiving temporal data to cloud based storage systems like s3, gcloud storage as temporal service account can be associated to IAM roles so that the temporal pod[s] can authenticate to S3/gcloud storage using service account 

* Support creating custom service account with a default to be "default" namespace (kubernetes default namespace)
    * Make creating custom service account OPTIONAL, with by default make it NOT to create it, for backward compatibility
* Assign the service account to all pods which are part of deployments and jobs, so that all pods of temporal will use same service accounts

hkumarmk:
Fixes: issue #245

## What was changed
This patch make the changes that need to fix the issue/feature request added in #245  :

* Add option in values.yaml that enable optionaly creating custom
  service account, with following properties:
  * values.yaml has set with serviceAccount.create to false, so by
    default they use kubernetes default service account - this will
    make this change backward compatible
  * Once serviceAccount.create is marked to true, it use default name
    which is temporal.fullname, but can be customized with serviceAccount.name
  * Users can add extraAnnotations to add any additional annotations for
    service acccounts
    * Usually associating k8s service account with public cloud IAM role
      is done by adding custom annotation to serviceAccount, so this
      will enable users to pass such annotations to temporal
      serviceAccount
* Created a custom serviceAccount with pre-install helm hook with low
  hook-weight as temporal server job is running with pre-install
  helm hook in some cases and serviceAccount should be created before
  job execution for the jobs run with custom serviceAccount
## Why?
Details added in issue #245 

## Checklist
1. Closes #245 

2. How was this tested:
    * In all scenarios below all the pods of deployments and jobs have serviceAcountName defined
    * Ran helm install with serviceAcccount.create=true in which case serviceAccount with name equal to temporal.fullname is created and is used for all pods
    * Ran helm install with serviceAcccount.create=true, serviceAccount.name="myserviceaccount" and serviceAccount with name "myserviceaccount" is created and is used for all pods
    * Ran helm install with serviceAcccount.create=true, and added serviceAccount.extraAnnotations which are key value pairs,  and erviceAccount with name equal to temporal.fullname is created and those annotations added in addition to other annotations defined in the templates and the service account is used for all pods
    * Ran helm install with serviceAcccount.create=false in which case  no service accounts are created and the serviceAccount with name "default", is used for all pods. NOTE that "default" is the default serviceAccount k8s assign all pods to.

3. Any docs updates needed?

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -32,6 +32,22 @@ Create chart name and version as used by the chart label.
 {{- end -}}
 
 {{/*
+Create the name of the service account
+*/}}
+{{- define "temporal.serviceAccountName" -}}
+{{ default (include "temporal.fullname" .) .Values.serviceAccount.name }}
+{{- end -}}
+
+{{/*
+Define the service account as needed
+*/}}
+{{- define "temporal.serviceAccount" -}}
+{{- if .Values.serviceAccount.create -}}
+serviceAccountName: {{ include "temporal.serviceAccountName" . }}
+{{- end -}}
+{{- end -}}
+
+{{/*
 Create a default fully qualified component name from the full app name and a component name.
 We truncate the full name at 63 - 1 (last dash) - len(component name) chars because some Kubernetes name fields are limited to this (by the DNS naming spec)
 and we want to make sure that the component is included in the name.
--- a/templates/admintools-deployment.yaml
+++ b/templates/admintools-deployment.yaml
@@ -28,6 +28,7 @@ spec:
         app.kubernetes.io/component: admintools
         app.kubernetes.io/part-of: {{ .Chart.Name }}
     spec:
+      {{ include "temporal.serviceAccount" . }}
       containers:
         - name: admin-tools
           image: "{{ .Values.admintools.image.repository }}:{{ .Values.admintools.image.tag }}"
--- a/templates/server-deployment.yaml
+++ b/templates/server-deployment.yaml
@@ -41,6 +41,7 @@ spec:
         {{- toYaml . | nindent 8 }}
         {{- end }}
     spec:
+      {{ include "temporal.serviceAccount" $ }}
       {{- if or $.Values.cassandra.enabled (or $.Values.elasticsearch.enabled $.Values.elasticsearch.external)}}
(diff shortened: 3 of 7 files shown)


templates/_helpers.tpl
```diff
@@ -31,6 +31,17 @@ Create chart name and version as used by the chart label.
...
 {{- end -}}
 
+{{/*
+Create the name of the service account
+*/}}
+{{- define "temporal.serviceAccountName" -}}
+{{- if .Values.serviceAccount.create -}}
+    {{ default (include "temporal.fullname" .) .Values.serviceAccount.name }}
+{{- else -}}
+    {{ default "default" .Values.serviceAccount.name }}
```
nvtkaszpir, line 41: in some kubernetes environments there will not be 'default' service account at all. So it is better to skip 'else' section

templates/admintools-deployment.yaml
```diff
@@ -28,6 +28,7 @@ spec:
         app.kubernetes.io/component: admintools
         app.kubernetes.io/part-of: {{ .Chart.Name }}
     spec:
+      serviceAccountName: {{ include "temporal.serviceAccountName" . }}
```
nvtkaszpir, line 31: this should optionally allow setting service account to nil/empty/unset.

templates/_helpers.tpl
hkumarmk: @nvtkaszpir  done

templates/admintools-deployment.yaml
hkumarmk: @nvtkaszpir  this is done - please take a look at it

hkumarmk:
@nvtkaszpir  Fixed the issues that you pointed out.

underrun (approved):

PR #247, opened 2021-11-27
#121 adds initial support to mysql

### LINKED ISSUE #121: Missing SQL store support
I have been trying to deploy Temporal.io with MySQL store support. I see the rest of the chart is heavily inspired from [banzaicloud's cadence chart](https://github.com/banzaicloud/banzai-charts/tree/master/cadence), but it seems MySQL support has been explicitly removed from the chart while Temporal itself seems to support it.

Is there a specific reason?

I have made changes to this chart to support it for my use for now. If the Temporal team/community is interested, I can create a PR.

hammett:
## What was changed
This is an initial, minimalistic support to use mysql as a store for temporal and visibility via the helm charts

## Why?
Currently the helm charts seems to only support cassandra out of the box.

## Checklist

1. Closes #121  which seems to be closed but the support isn't there

2. How was this tested:
This was tested locally with minikube

3. Any docs updates needed?
What needs to be stated is that this brings mysql within the temporal chart the same way it's doing for cassandra. Next step is making it work with an externally configured instance.

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -36,6 +36,10 @@ dependencies:
     repository: https://grafana.github.io/helm-charts
     version: 5.0.10
     condition: grafana.enabled
+  - name: mysql
+    version: '8.8.1'
+    repository: https://charts.bitnami.com/bitnami
+    condition: mysql.enabled
 
 # A chart can be either an 'application' or a 'library' chart.
 #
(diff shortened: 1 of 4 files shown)


underrun:
we will not be adding more dependencies to this chart - if you would like to run temporal using mysql this is already possible by setting up a mysql server in another way (which could be with a helm chart) and then deploying temporal

PR #249, opened 2021-12-11
allow for sql db setup/update

plytro:
## What was changed
This allows an external sql database to be setup and updated with the chart like Cassandra can be. Using the follow values set, a functioning external sql database is seeded with data. 

```
schema:
  setup:
    enabled: true
  update:
    enabled: true

server:
  config:
    persistence:
      default:
        driver: sql
        sql:
          driver: mysql
          host: dbhost
          port: 3306
          database: temporal
          user: myuser
          password: mypassword
      visibility:
        driver: sql
        sql:
          driver: mysql
          host: dbhost
          port: 3306
          database: temporal_visibility
          user: myuser
          password: mypassword
```

## Why?
When installing into kubernetes using an sql server, one shouldn't be required to download and run the sql-setup tool to get a functioning db backend.

## Checklist

1. How was this tested:
Using the above persistence configuration the chart successfully installed, created the database in mysql, setup the initial 0.0 schema, and then upgraded to the latest schema for the container image used.

3. Any docs updates needed?
README.md has been updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,37 @@ Example:
 
 You might already be operating a MySQL instance that you want to use with Temporal.
 
+#### Configuring using the chart
+
+In this case, create your database host and user with a permissions grant that allows the user to create databases and execute database schema changes.
+
+Once you initialized the database server, fill in the configuration values in `values/values.mysql.yaml`, change `schema.setup.enabled: true`, change `schema.update.enabled: true`
+
+*NOTE:* For MYSQL <5.7.20 (e.g AWS Aurora MySQL) use `values/values.aurora-mysql.yaml`
+
+```bash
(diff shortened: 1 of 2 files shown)


sw-yx:
appreciate this PR, but unfortunately is off topic for the goals of this repo. However, we're trying to preserve the modifications people make for others to find, so have added an `Alternative` label that we'll be promoting more in coming months as we curate good modifications.

jcgarciam-signavio:
@sw-yx sorry for commenting on a closed PR, however I find your reason not to be valid. When the helm chart is being used with cassandra the schema get automatically updated, but when people want to uses either mysql or postgresql they are left with an incomplete setup which require additional preparation.

If this helm chart wants to support both cassandra and sql environment, it should make it in a consistent way.

PR #257, opened 2022-01-19
accept config.global

wenerme:
accept other global config like auth


sw-yx:
thank you for the other PR, but this one we dont think is as useful for a general audience.

PR #265, opened 2022-01-26
Render schema-setup job's init-containers when using external Cassandra cluster

### LINKED ISSUE #264: [Bug] Jobs do not render create-default-store and create-visibility-store init-containers for externally provided Cassandra

I am using the chart to initially set up a Temporal instance on an existing cassandra cluster by setting `cassandra.enabled=false` and `schema.setup.enabled=true`.

### Describe the bug

With both above conditions set, the `schema-setup` job will not create init-containers `create-default-store
` and `create-visibility-store` and subsequently fails with an error: `Keyspace 'temporal' does not exist` during the container run.

Example:

```
      initContainers:
          []
      containers:
        - name: default-schema
```

Expected output should include rendered init-containers:

```
      initContainers:
        - name: create-default-store
          (... truncated)
        - name: create-visibility-store
          (... truncated)
      containers:
        - name: default-schema
```

### Minimal Reproduction

Check out the master branch of this repository and run:

```
helm template temporal . --values values/values.cassandra.yaml --set schema.setup.enabled=true  
```

### Environment/Versions

- Latest helm chart master branch commit [240cc89](https://github.com/temporalio/helm-charts/tree/240cc8909d0da9a26aebab33fb8ef67949b89fa0)
- Helm v3.7.2

perplexa:
## What was changed

This PR fixes the bug described in issue #264 

## Why?

The `schema-setup` job will not render its `init-containers` when provided with an external Cassandra cluster (`cassandra.enable=false` set in the chart)

## Checklist

1. Closes #264

2. How was this tested:

```
helm template temporal . --values values/values.cassandra.yaml --set schema.setup.enabled=true
```

Before the fix, init-containers are missing, after the fix they are correctly rendered.

3. Any docs updates needed?

- none

### DIFF
--- a/templates/server-job.yaml
+++ b/templates/server-job.yaml
@@ -39,7 +39,9 @@ spec:
       {{ include "temporal.serviceAccount" . }}
       restartPolicy: "OnFailure"
       initContainers:
-        {{- if or .Values.cassandra.enabled }}
+        {{- if or .Values.cassandra.enabled
+                  (eq (include "temporal.persistence.driver" (list $ "default")) "cassandra")
+                  (eq (include "temporal.persistence.driver" (list $ "visibility")) "cassandra") }}
         {{- if .Values.cassandra.enabled }}
         - name: check-cassandra-service
           image: busybox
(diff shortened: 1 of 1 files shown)


underrun:
this is interesting - i'm not sure we want to run init-containers if using an external database. mostly this stuff is only for dev and the recommendation would be if you're using an external database to use the `temporal-cassandra-tool` to configure the clusters for temporal.

thoughts?

perplexa:
I am trying to run `temporal-cassandra-tool` through the init-container :)

I think it's fair that if somebody does not want to run the schema setup through the chart, they can set `schema.setup.enabled=false` to deactivate the entire job that runs this. However, if enabled without this fix, the job will directly jump to `temporal-cassandra-tool setup-schema` and skip `temporal-cassandra-tool create` and therefore fail.

My current situation is that the Cassandra chart that's shipping with Temporal's chart does not start successfully and I have set up Cassandra using the bitnami chart (also for a non-production setup) but would still like to be able to use the schema creation job of this chart successfully.

wdyt?

underrun:
yeah - i get it now. i think my preference would have been to only allow either create keyspace or update of schema if the internal cassandra were used. but i agree that the current chart doesn't really lean that direction. i'm okay with this path forward to fix the issue you're running into.

underrun (approved):

Issue #268, opened 2022-02-08
[Feature Request] Add the ability to set custom secret key for existing persistence secret


JoshVee:
### Is your feature request related to a problem? Please describe.

The Helm chart currently defaults to a secret key `password` when using an `existingSecret` for persistence credentials.
This does not work in all circumstances, for example the [Bitnami Postgres chart creates a secret](https://github.com/bitnami/charts/blob/master/bitnami/postgresql-ha/templates/postgresql/secrets.yaml#L19) with `postgresql-password`.

### Describe the solution you'd like

The chart could provide a dynamic value for the `existingSecretKey` and default to `password` if it's not set.

### Additional context

Example error when using the Bitnami chart:

```
containerStatuses:
    - image: docker.io/temporalio/server:1.14.0
      imageID: ''
      lastState: {}
      name: temporal-worker
      ready: false
      restartCount: 0
      started: false
      state:
        waiting:
          message: >-
            couldn't find key password in Secret
            default/app-staging-postgres-postgresql
          reason: CreateContainerConfigError
```

joebowbeer:
The secretKey is derived here:

https://github.com/temporalio/helm-charts/blob/master/templates/_helpers.tpl#L314

kenske:
@joebowbeer I believe that's a bug, instead of 
```
{{- print "password" -}}
```

it should read

```
{{- $storeConfig.sql.secretKey -}}
```

robholland:
This should now be fixed, we refer to `secretKey` in the values for the driver.

PR #269, opened 2022-02-08
feat: add the ability to customise existing secret key

JoshVee:
## What was changed

Add a custom chart value for `existingSecretKey`.

## Why?

Adds more flexibility when deploying databases using unrelated charts.

## Checklist

1. Closes https://github.com/temporalio/helm-charts/issues/268

2. How was this tested:

Manual testing via Helm CLI

3. Any docs updates needed?

### DIFF
--- a/templates/_helpers.tpl
+++ b/templates/_helpers.tpl
@@ -203,7 +203,7 @@ Source: https://stackoverflow.com/a/52024583/3027614
 {{- $store := index . 1 -}}
 {{- $storeConfig := index $global.Values.server.config.persistence $store -}}
 {{/* Cassandra password is optional, but we will create an empty secret for it */}}
-{{- print "password" -}}
+{{- default "password" $storeConfig.cassandra.existingSecretKey -}}
 {{- end -}}
 
 {{- define "temporal.persistence.sql.driver" -}}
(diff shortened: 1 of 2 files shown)


JoshVee:
@alexshtin is this PR okay for review?

templates/_helpers.tpl
```diff
@@ -203,7 +203,7 @@ Source: https://stackoverflow.com/a/52024583/3027614
 {{- $store := index . 1 -}}
 {{- $storeConfig := index $global.Values.server.config.persistence $store -}}
 {{/* Cassandra password is optional, but we will create an empty secret for it */}}
-{{- print "password" -}}
+{{- default "password" $storeConfig.cassandra.existingSecretKey -}}
```
robholland, line 206: This syntax is backwards, the default value comes second.

robholland (changes requested):

robholland:
The `secretKey` field will be used when using an existing secret, no need for another field here.

templates/_helpers.tpl
JoshVee: No, it doesn't.

http://masterminds.github.io/sprig/defaults.html

robholland:
Sorry, I mixed dockerize with Helm, you are correct.

PR #276, opened 2022-02-28
default namespace setup

everesio:
## What was changed
Optional job to create a default namespace.

## Why?
Current deployment needs a manual step 

## Checklist

1. Closes 

2. How was this tested:
Tested on k8s with postgresql DB created by zalando postgres operator

3. Any docs updates needed?

### DIFF
--- a/templates/server-job.yaml
+++ b/templates/server-job.yaml
@@ -301,3 +301,64 @@ spec:
       {{- end }}
 {{- end }}
 {{- end }}
+
+{{- if .Values.defaultnamespace.setup.enabled }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ include "temporal.componentname" (list . "default-ns-setup") }}
+  labels:
+    app.kubernetes.io/name: {{ include "temporal.name" . }}
+    helm.sh/chart: {{ include "temporal.chart" . }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/version: {{ .Chart.AppVersion | replace "+" "_" }}
+    app.kubernetes.io/component: auto-setup
+    app.kubernetes.io/part-of: {{ .Chart.Name }}
+  annotations:
+    "helm.sh/hook": post-install,post-upgrade
+    "helm.sh/hook-weight": "2"
+    {{- if not .Values.debug }}
+    "helm.sh/hook-delete-policy": hook-succeeded,hook-failed
+    {{- end }}
+spec:
+  backoffLimit: {{ .Values.defaultnamespace.setup.backoffLimit }}
+  template:
+    metadata:
+      name: {{ include "temporal.componentname" (list . "default-ns-setup") }}
+      labels:
+        app.kubernetes.io/name: {{ include "temporal.name" . }}
+        helm.sh/chart: {{ include "temporal.chart" . }}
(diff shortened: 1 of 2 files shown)


bergundy (commented):
Why not let the helm chart register any number of arbitrary namespaces?
It’s more flexible, not called “default” (which may be confused with the name default) and doesn’t need an `enabled` flag.

templates/server-job.yaml
```diff
@@ -301,3 +301,64 @@ spec:
...
+      containers:
+        - name: create-default-namespace
+          image: "{{ $.Values.admintools.image.repository }}:{{ $.Values.admintools.image.tag }}"
+          imagePullPolicy: {{ $.Values.admintools.image.pullPolicy }}
+          command: ['sh', '-c', 'tctl --ns $(DEFAULT_NAMESPACE) namespace describe || tctl --ns $(DEFAULT_NAMESPACE) namespace register --rd $(DEFAULT_NAMESPACE_RETENTION) --desc "Default namespace for Temporal Server."']
+          env:
+            - name: DEFAULT_NAMESPACE
+              value: "{{ .Values.defaultnamespace.setup.name }}"
+            - name: DEFAULT_NAMESPACE_RETENTION
+              value: "{{ .Values.defaultnamespace.setup.retention }}"
```
bergundy, line 349: This doesn’t support tls afaict, which is a fairly common requirement when setting up the server.

templates/server-job.yaml
```diff
@@ -301,3 +301,64 @@ spec:
...
+        app.kubernetes.io/component: auto-setup
+        app.kubernetes.io/part-of: {{ .Chart.Name }}
+    spec:
+      {{ include "temporal.serviceAccount" . }}
+      restartPolicy: "OnFailure"
+      containers:
+        - name: create-default-namespace
+          image: "{{ $.Values.admintools.image.repository }}:{{ $.Values.admintools.image.tag }}"
+          imagePullPolicy: {{ $.Values.admintools.image.pullPolicy }}
+          command: ['sh', '-c', 'tctl --ns $(DEFAULT_NAMESPACE) namespace describe || tctl --ns $(DEFAULT_NAMESPACE) namespace register --rd $(DEFAULT_NAMESPACE_RETENTION) --desc "Default namespace for Temporal Server."']
```
bergundy, line 344: Can you please use temporal cli instead of the soon to be deprecated tctl?

bergundy: I think this command is missing quotes around the variables

nikanmog (approved):

templates/server-job.yaml
```diff
@@ -301,3 +301,64 @@ spec:
       {{- end }}
 {{- end }}
 {{- end }}
```
debugger24, line 303:
```suggestion
{{- end }}
---
{{- end }}
```

values.yaml
```diff
@@ -405,3 +405,10 @@ cassandra:
 
 mysql:
   enabled: false
+
+defaultnamespace:
+  setup:
+    enabled: false
+    name: default
+    backoffLimit: 100
+    retention: 1
```
fm2022aa, line 414: use bigger default value?

values.yaml
```diff
@@ -405,3 +405,10 @@ cassandra:
 
 mysql:
   enabled: false
+
+defaultnamespace:
+  setup:
+    enabled: false
+    name: default
```
fm2022aa, line 412: can support multi name?

robholland:
Implemented in https://github.com/temporalio/helm-charts/pull/550

PR #280, opened 2022-03-18
add custom pod label options to chart

deltamualpha:
This is a relatively minor quality-of-life change that I asked about in Slack and got a positive response about.

## What was changed
This PR adds the ability to add custom labels to the deployment specs, so that pods all launch with them. It supports adding labels to the pods in the cluster itself (by service type, as well as globally), the web UI pods, and the admintools pod.

## Why?
In my case, we control how log output is parsed through pod labels, and adding labels to tell our log aggregator to parse temporal's logs as JSON was a required step. Other use-cases might be adding tags from deployment systems noting who triggered the deployment, or other bookkeeping tasks.

## Checklist
2. How was this tested:
I ran this PR through `helm template -f values.yaml .` and confirmed no change to the output from previous, then added some label key:value pairs and confirmed the appeared in the correct places in the YAML.

3. Any docs updates needed?
As far as I can tell, the existing custom annotation functionality is undocumented; custom pod labels in value.yaml files seems to be a very common idiom across helm charts, so I'm not sure there's much temporal-specific documentation required.

### DIFF
--- a/templates/admintools-deployment.yaml
+++ b/templates/admintools-deployment.yaml
@@ -28,6 +28,9 @@ spec:
         app.kubernetes.io/version: {{ .Chart.AppVersion | replace "+" "_" }}
         app.kubernetes.io/component: admintools
         app.kubernetes.io/part-of: {{ .Chart.Name }}
+        {{- with $.Values.admintools.podLabels }}
+        {{- toYaml . | nindent 8 }}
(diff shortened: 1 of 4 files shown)


underrun:
Thanks for the PR! Will take a look and get back to you soon.

underrun (approved):

PR #299, opened 2022-06-08
feat: Enable LivenessProbe for workers

### LINKED ISSUE #298: [Feature Request] Enable LivenessProbe for workers

When we deploy a new cluster, the workers nodes don't have a LivenessProbe because it's explicit configured in the code to not add them.

Remove the if loop in the lines above. Or implement a specific livenessProbe for workers.

https://github.com/temporalio/helm-charts/blob/master/templates/server-deployment.yaml#L127-L132
```
{{- if ne $service "worker"}}
          livenessProbe:
             initialDelaySeconds: 150
             tcpSocket:
               port: rpc
 {{- end }}
```

gracco:
## What was changed
This pull request removes the loop to not add livenessProbe for workers

## Why?
Every service should have a livenessProbe, in case of failure the kubelet be capable to restart the pod.

## Checklist

1. Closes #298

2. How was this tested:
It was applied via helm in our cluster.

3. Any docs updates needed?
No

### DIFF
--- a/templates/server-deployment.yaml
+++ b/templates/server-deployment.yaml
@@ -124,12 +124,10 @@ spec:
             - name: metrics
               containerPort: 9090
               protocol: TCP
-          {{- if ne $service "worker"}}
           livenessProbe:
              initialDelaySeconds: 150
              tcpSocket:
                port: rpc
-          {{- end }}
           volumeMounts:
             - name: config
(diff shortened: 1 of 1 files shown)


robholland:
The worker does not have a port we can use to health check.

Issue #313, opened 2022-08-19
[Feature Request] Turn repo to a helm repo


amirsaber:
### Is your feature request related to a problem? Please describe.

We are trying to add this repository as a helm repository for our local development cluster but it's failing to detect repo due to missing index.yaml

```
Failed  25s (x10 over 13m)  source-controller  failed to fetch Helm repository index: failed to cache index to temporary file: failed to fetch https://github.com/temporalio/helm-charts/index.yaml : 404 Not Found  
```

### Describe the solution you'd like

A workflow could be added to this repo to maintain index.yaml as Github pages as described [here](https://helm.sh/docs/howto/chart_releaser_action/#github-actions-workflow)

adamstrawson:
It looks like you're using Flux based on your log (`source-controller`); 

To use a Git Repository, you can use the `GitRepository` kind. 

For example:

```
---
apiVersion: source.toolkit.fluxcd.io/v1beta1
kind: GitRepository
metadata:
  name: temporal
spec:
  interval: 1m
  url: https://github.com/temporalio/helm-charts
  ref:
    tag: v1.17.1
```

amirsaber:
Thanks for the pointer @adamstrawson This will remove the need for the PR however we should still stick to our own repo since the helm charts are not starting successfully due to https://github.com/temporalio/helm-charts/issues/182

amirsaber:
Closing this issue since no change is required

PR #314, opened 2022-08-19
Restructure to support helm chart releaser action

amirsaber:
## What was changed

Restructured the chart files into chart folder and created a helm releaser workflow

## Why?
So that helm releaser will pick the chart and create github pages branch in order to be able to use the repository as helm repository

## Checklist

1. Closes https://github.com/temporalio/helm-charts/issues/313

2. How was this tested:
Tested in a local k8s using forked repo

3. Any docs updates needed?
Updated README file to match new folder structure


amirsaber:
Closing PR since a workaround is mentioned in the issue

PR #322, opened 2022-10-04
fix(doc): update text around creating/setup of DB

skandragon:
## What was changed

Add comments to update the current create-database syntax (from main), and add examples for PostgreSQL to create them with restricted user.

## Why?

It looks like the command has changed in the temporal repo.

3. Any docs updates needed?

It's all about docs.  :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,11 +134,11 @@ export SQL_PORT=3306
 export SQL_USER=mysql_user
 export SQL_PASSWORD=mysql_password
 
-./temporal-sql-tool create-database -database temporal
+./temporal-sql-tool create-database temporal
 SQL_DATABASE=temporal ./temporal-sql-tool setup-schema -v 0.0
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -171,6 +171,14 @@ You might already be operating a PostgreSQL instance that you want to use with T
 
 In this case, create and configure temporal databases on your PostgreSQL host with `temporal-sql-tool`. The tool is part of [temporal repo](https://github.com/temporalio/temporal), and it relies on the schema definition, in the same repo.
 
+If the user temporal will connect with does not have permissions to create the databases, create them and skip the
+`create-database` commands.  For example, with PostgreSQL, one could:
```
mindaugasrukas, line 175:
```suggestion
If a user does not have permission to create the databases, create them and skip the `create-database` commands. For example, with PostgreSQL, one could:
```

Issue #319, opened 2022-09-15
Docker image tag for Web component is not pinned


thjwhite:
### Is your feature request related to a problem? Please describe.

In a recent change for the web UI component it looks like the docker image tag was changed to be "latest" rather than a pinned docker tag. Unfortunately this brings in an opportunity for unexpected behavior / bugs without an explicit deployment. For example a pod may get deleted and pull in a newer version which may change behavior or have bugs. I am not sure if there was any reason for the latest tag so wanted to bring it up to the team.

### Describe the solution you'd like

Change the values.yaml "web.image.tag" to be a pinned version.

### Additional context

This morning we ran into the issue referenced in this linked patch release relating to CSS/JS files not loading. If the versions were pinned it would greatly help us reduce unexpected changes in the UI component of Temporal and allow us to upgrade when we have time to validate the deployment.
https://github.com/temporalio/ui-server/releases/tag/v2.6.1

deltamualpha:
Agreed with the need to pin the ui container tag here. Even setting aside the unexpected-upgrades point, if you change `replicaCount` to anything > 1, new k8s nodes that have the UI container scheduled onto them will pull the newest version of the ui-server container (because of `pullPolicy: IfNotPresent`), which serves different assets from existing running containers. This immediately breaks the UI entirely, as assets are not stable across releases.

robholland:
Long since fixed.

PR #337, opened 2022-12-06
setting specific img tag for temporalio/ui

tsurdilo:
https://github.com/temporalio/helm-charts/issues/319

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -266,7 +266,7 @@ web:
 
   image:
     repository: temporalio/ui
-    tag: latest
+    tag: 2.9.0
     pullPolicy: IfNotPresent
 
   service:


values.yaml
```diff
@@ -266,7 +266,7 @@ web:
 
   image:
     repository: temporalio/ui
-    tag: latest
+    tag: 2.9.0
```
yiminc, line 269: why this change? it is hard to keep this in sync, and we will have to upgrade this repo whenever ui upgrade.

tsurdilo: users would need to update this similar to like grafana and prometheus image tags.
this is just so that "latest" is not used and have certain pods be able to pull a newer image tag by mistake

yiminc (approved):

PR #347, opened 2022-12-16
add hints for prod deployments

Hades32:
## What was changed
This adds more example values to the postgres values file

## Why?
It took me far too long to figure out these values as the aren't documented and I had look through the source code

## Checklist
2. How was this tested:
It's only commented out/already set default values. But when commenting those then, that's what we use for our deployments

3. Any docs updates needed?
Maybe it would make sense to add this to the docs too, but they are already pretty long and the point you to this file anyway

### DIFF
--- a/values/values.postgresql.yaml
+++ b/values/values.postgresql.yaml
@@ -11,8 +11,15 @@ server:
           database: temporal
           user: _USERNAME_
           password: _PASSWORD_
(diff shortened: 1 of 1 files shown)


mindaugasrukas (approved):
LGTM
@underrun any comments?

underrun (approved):

PR #281, opened 2022-03-21
Automatically configure Postgres and MySQL databases on startup.

rogersd:
## What was changed

Automatically create databases and apply the schemas to them for Postgres and MySQL.

## Why?

In order to be able to create new environments via Terraform without having to manually create the databases between the provisioning of the cluster and the deployment of the helm chart.

Closes #349

## Checklist

1. How was this tested:

I used this to provision a Postgres cluster, but no other testing has been done.  In particular, MySQL hasn't been tested at all.

2. Any docs updates needed?

Probably?  I'm just giving this back to hopefully help other people, but I don't have the time to do any more work on it myself.

PR #350, opened 2023-01-06
Add PostgreSQL schema jobs

### LINKED ISSUE #349: [Feature Request] Add Kubernetes job to migrate postgresql schema
Currently, the helm chart only supports migrating schemas on Cassandra; I would like to add support for the same feature when bringing my own PostgreSQL

Add support for PostgreSQL for the following values in `values.yaml`

```
schema:
  setup:
    enabled: true
  update:
    enabled: true
```

dkravetz:
## What was changed
Add kubernetes job to support database schema migration and updates on PostgreSQL. Also updated `values.postgresql.yaml` to reflect this 

## Why?
Current workplace uses a Bring Your Own PostgreSQL and having this change upstream would be simple for more people.

## Checklist

1. Closes #349 

2. How was this tested:
Set up a brand new PostgreSQL; provide hostname, username, and password, and set enable 

```
schema:
  setup:
    enabled: true
  update:
    enabled: true
```

3. Any docs updates needed?
No

### DIFF
--- a/values/values.postgresql.yaml
+++ b/values/values.postgresql.yaml
@@ -35,9 +35,13 @@ mysql:
 
 postgresql:
   enabled: true
+  image:
+    repo: postgres
+    tag: 15.1
+    pullPolicy: IfNotPresent
 
 schema:
   setup:
-    enabled: false
+    enabled: true
   update:
-    enabled: false
+    enabled: true
--- a/values/values.postgresql.yaml
+++ b/values/values.postgresql.yaml
@@ -6,11 +6,11 @@ server:
 
         sql:
           driver: "postgres"
-          host: _HOST_
+          host: "rokerage-dev-db-temporal.cluster-cy2utjok3e0t.eu-central-1.rds.amazonaws.com"
           port: 5432
           database: temporal
-          user: _USERNAME_
-          password: _PASSWORD_
+          user: master
+          password: "KkLyDxZmHZssEwCtpE6D3mWpIilQL3qE"
           maxConns: 20
           maxConnLifetime: "1h"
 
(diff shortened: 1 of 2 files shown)


medamineamara:
I have been working with this feature 

But I had to fix some bugs

- The  `PORT` as an env should be quoted.
- The cassandra Job should be disabled when the driver is postgres.
- Update job should wait for create job .. so I passed extra arg to pg_isready to verify the DB (--dbname).
- DB names were hardcoded but they should be retrieved from the values.
- used `temporal.persistence.schema` in `_helpers.tpl` to set schema directory name.
- removed the replace '_' by '-' in the DB name when updating schema.

I made a branch with this feature including my changes
https://github.com/medamineamara/Temporal-helm-charts/tree/fix-postgresql-schema

This commit contains all the listed changes 
https://github.com/medamineamara/Temporal-helm-charts/commit/1f9863778d53ff70fa350cbec4d717300378cf62

Should I open a new Pull Request ?

brandonros:
@medamineamara can you update your branch? it's a bit behind master and open a PR please? thank you so much for this!

values/values.postgresql.yaml
```diff
@@ -6,11 +6,11 @@ server:
 
         sql:
           driver: "postgres"
-          host: _HOST_
+          host: "rokerage-dev-db-temporal.cluster-cy2utjok3e0t.eu-central-1.rds.amazonaws.com"
```
Eusebiotrigo, line 9: I would remove these values, host, user and password.

brandonros:
this should probably be closed in favor https://github.com/temporalio/helm-charts/pull/281 i'm guessing

peter-c-larsson:
Think this one can be closed now, or?

Was about to start similar work myself but no point in case there is already open pull request. The referenced issue #281 has been closed in favour of #405 which is merged but there is no documentation updates making it a little bit tricky and not sure it actually works for our use case where we use the gcp auth proxy sidecar.

robholland:
Resolved via #405

Issue #343, opened 2022-12-09
Add GitHub Action linter job


mindaugasrukas:
### Is your feature request related to a problem? Please describe.

No checks on PRs.

### Describe the solution you'd like

Possible solution: https://github.com/helm/chart-testing-action

### Additional context

None

robholland:
Closing in favour of PR.

PR #352, opened 2023-01-22
WIP: Adding Linting and Testing on all committs

edmondop:
## What was changed

Addresses #343  

Getting the build to pass seems to be blocked by https://github.com/temporalio/helm-charts/issues/182 but I don't think disabling hooks is appropriate:

https://github.com/temporalio/helm-charts/compare/master...BAXUSNFT:temporal-helm-charts:master#diff-5f8eb04d49dd7caffd33da659dd7bae84435c114a6255319b2965dcf6a2536ab

As you see from my fork https://github.com/edmondop/helm-charts/actions/runs/3988830053/jobs/6840499585 , the install fails because the default keyspace is never ready.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: azure/setup-helm@v1
+        with:
+          version: v3.8.1
+
+      # Python is required because `ct lint` runs Yamale (https://github.com/23andMe/Yamale) and
+      # yamllint (https://github.com/adrienverge/yamllint) which require Python
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
```
underrun, line 24: does this require a really old version of python? 3.7 will be EOL in a few months.

edmondop: 👍 will update

.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
+name: Lint and Test Charts
+
+on: push
+
+jobs:
+  lint-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
```
mindaugasrukas, line 10: Let's use the `v3` version.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          fetch-depth: "0"
+
+      - name: Set up Helm
+        uses: azure/setup-helm@v1
```
mindaugasrukas, line 15: The latest version is `v3`. Any reason for using `v1`?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+      # yamllint (https://github.com/adrienverge/yamllint) which require Python
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+
+      - name: Set up chart-testing
+        uses: helm/chart-testing-action@v2.2.1
+        with:
+          version: v3.5.1
```
mindaugasrukas, line 29: These are also old versions, any reason for not using the latest?

ci/kind-test-values.yaml
```diff
@@ -0,0 +1,8 @@
+elasticsearch:
```
mindaugasrukas, line 1: Where is this file used?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+
+      - name: Run chart-testing (lint)
+        run: ct lint
+
+      - name: Increase inotify limits
+        run: |
+          echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sysctl -p
+          echo fs.inotify.max_user_instances=512 | sudo tee -a /etc/sysctl.conf && sysctl -p
```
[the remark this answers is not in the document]
edmondop: It was an effort to try to get the helm chart install on a single node kind, after seeing some errors in the cassandra logs. It turns out that it dosn't make a difference

ci/kind-test-values.yaml
edmondop: From https://github.com/helm/chart-testing/blob/main/doc/ct_install.md

> Charts may have multiple custom values files matching the glob pattern '*-values.yaml' in a directory named 'ci' in the root of the chart's directory. The chart is installed and tested for each of these files. If no custom values file is present, the chart is installed and tested with defaults.

underrun:
I'd love to land this - but i'm inclined to skip the install step if it's not completing unless/until the deadlock gets sorted out

edmondop:
Sure, I put it on hold because I was trying to get an hint on why this would not work on a kind cluster.

In the meanwhile I have realised that GitHub has added large runners, so I would give a shot using those, what do you think? Can anyone set up a runner group ? https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+on: pull_request
+
+jobs:
+  lint-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: "0"
```
nvtkaszpir, line 12: I suggest keeping it to 50, if MR has more 50 commits it should be squashed then.

ct.yml
```diff
@@ -0,0 +1,6 @@
+# chart-repos:
+#   - cassandra=https://charts.helm.sh/incubator
+#   - prometheus=https://prometheus-community.github.io/helm-charts
+#   - elasticsearch=https://helm.elastic.co
+#   - grafana=https://grafana.github.io/helm-charts 
+validate-maintainers: false
```
nvtkaszpir, line 6: please fix new line at the end of the file

nvtkaszpir (changes requested):

.github/workflows/ci.yml
edmondop: Fixed with https://github.com/temporalio/helm-charts/pull/352/commits/5190b68e1972ada4cdb405e5f0bfbd03f801e0ca

ct.yml
edmondop: Fixed with https://github.com/temporalio/helm-charts/pull/352/commits/5190b68e1972ada4cdb405e5f0bfbd03f801e0ca

robholland:
To make some iterative progress here, can you please re-base and get lint-only running? Then we can work on install checks after.

robholland:
I will work with you to ensure this gets merged.

robholland (changes requested):
Re-base on current code, just lint for now.

edmondop:
> Re-base on current code, just lint for now.

Thanks, on it. Should be able to get it ready by eod

.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
...
+
+      - name: Set up chart-testing
+        uses: helm/chart-testing-action@v2.5.0
+        with:
+          version: v3.8.0
+
+      - name: Run chart-testing (lint)
+        run: ct lint --charts ./ --config=ct.yml
+                
+      - name: Create kind cluster
```
robholland, line 34: We don't need the cluster for now as we're not testing install.

robholland (changes requested):
Please add a `ci/` values file that tests cassandra + ElasticSearch combination (we mention that as our "default" in the README).

charts/temporal/templates/server-deployment.yaml
```diff
@@ -1,6 +1,6 @@
 {{- if $.Values.server.enabled }}
 {{- range $service := (list "frontend" "history" "matching" "worker") }}
-{{ $serviceValues := index $.Values.server $service }}
+{{- $serviceValues := index $.Values.server $service }}
```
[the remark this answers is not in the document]
edmondop: No it might be result of a bad rebase:)

.github/workflows/ci.yml
edmondop: Do you suggest we add elasticsearch + cassandra but not run the tests in CI? Also I discovered I had to upgrade the action and now it doesn't work anymore :(

This is the last build with 2.4.0 that now fails because Google deprecated some APIs
https://github.com/edmondop/helm-charts/actions/runs/9551467429/job/26325831512

using the new version fails for other reasons , see https://github.com/edmondop/helm-charts/actions/runs/9551593568/job/26326271659#step:6:21. Maybe they changed the expected directory struxture, will investigate. If you are on the Slack channel you can ping me there so we can close this

robholland: Did ping on slack but haven't heard back. I suggested adding a new values file for the cassandra/ES case, but now that I think about it, maybe the values files doesn't matter for lint? And no, we shouldn't do the install tests yet, so don't need the Kind cluster.

Issue #353, opened 2023-01-29
[Bug] Temporal services are not starting when deployed in istio enabled namespace


jeenadeepak:
### What are you really trying to do?
I am running temporal with ISTIO enabled namespace as we are using istio mesh.

### Describe the bug
Temporal run schema setup job before deploying temporal cluster, and expect that these startup job should be successful executed, but in istio enabled namespace istio run as sidecar due to which these job not terminate successfully and deployment get stuck in between.

### Minimal Reproduction
Install istio and enable istio-injection="enabled" for temporal namespace.

### Environment/Versions

- OS and processor: EKS managed cluster
- Temporal Version: latest master branch
- Are you using Docker or Kubernetes or building Temporal from source?
No, I am installing using helm-chart

mindaugasrukas:
Related to #373

PR #354, opened 2023-01-29
Add support for ISTIO sidecar

jeenadeepak:
## What was changed

Add support for ISTIO deployment

## Why?

Istio run with each pod due to which one time job/pod are not successfully terminal, istio is not required here as db schema setup and update are one time job, hence skipping istio for jobs.

## Checklist

1. Closes 
 #https://github.com/temporalio/helm-charts/issues/353

2. How was this tested:

Testing on AWS EKS with istio.

3. Any docs updates needed?

no

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,4 @@
+# Syntax is here:
+# https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
+
+* @temporalio/server @temporalio/infra
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -49,8 +49,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.27.0
+version: 0.28.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 1.21.5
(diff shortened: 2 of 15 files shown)


mindaugasrukas:
Please make these optional via `extraAnnotations` value.
As an example, see https://github.com/temporalio/helm-charts/blob/master/templates/serviceaccount.yaml#L16.

mindaugasrukas (changes requested):
Please make these optional via `extraAnnotations` value.
As an example, see https://github.com/temporalio/helm-charts/blob/master/templates/serviceaccount.yaml#L16.

jeenadeepak:
@mindaugasrukas : I have checked this example, it does not seems correct solution.

I am facing this issue with DB migration job due to istio, rest service are working fine with istio, so I do not want remove istio for entire application, but as per this example this will skip istio for other temporal services.

There are 2 problem I have identified with DB migration:
1: DB migration is one time job, hence db migration container quit but istio container is log running so it showing error, 
2: DB migration started before istio networking hence it stop DB migration container to connect with SQL port.

mindaugasrukas:
My point is that this change will add environment-specific fields to the generic template. We don't want to add every environment-specific setting to the template but have the ability to inject user-specific configs. So, I'd like to ask you to make the template change, allowing you or anyone else to inject their needs.

vishwa-trulioo:
Any update on this?

PR #365, opened 2023-02-27
make chart more configurable

monotek:
## What was changed

* make securityContexts for pod & container configurable everywhere
* add resources config where not possible yet (schema & server-job)
* added configurable topologySpreadConstraints
* added configurable podDisruptionBudget for server components
* removed some blank lines from rendered resources
* current default settings are not changed

## Why?

* some resources not configurable which is problematic in clusters which enforce some of these settings via policies

## Checklist

1. Closes 

2. How was this tested:

`helm template .`

3. Any docs updates needed?

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -49,7 +49,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.20.0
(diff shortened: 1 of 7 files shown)


underrun (approved):
LGTM

mindaugasrukas (approved):

Issue #380, opened 2023-05-04
[Bug] helm-template fails when enabling serviceAccount


crhuber:
### What are you really trying to do?

I am trying to get the server pods to use a serviceAccount. 

### Describe the bug

In the values file I added the following with no other changes

```
serviceAccount:
  create: true
```

And I get the following error after running helm template

helm template temporal-server -f values.yaml .

> Error: YAML parse error on temporal/templates/server-deployment.yaml: error converting YAML to JSON: yaml: line 36: mapping values are not allowed in this context

### Minimal Reproduction

```
git clone https://github.com/temporalio/helm-charts
cd helm-charts/
vi values.yaml 
# add serviceAccount.create=true
helm dependency build
helm template temporal-server -f values.yaml .
```

returns:
> Error: YAML parse error on temporal/templates/server-deployment.yaml: error converting YAML to JSON: yaml: line 36: mapping values are not allowed in this context

### Environment/Versions

- OS and processor: m1 mac
- Temporal Version: temporal chart v1.20.2
- Helm version:  3.11.3

PR #381, opened 2023-05-10
fix: disable whitespace removal for include temporal.serviceAccount

2hamed:
Fixes issue: https://github.com/temporalio/helm-charts/issues/380

## What was changed
Removed the hyphen (`-`) at the beginning of the `include` function. It was causing all whitespaces to be removed from the output resulting in `serviceAccount: ...` to be on the same line as the previous statement.


crhuber:
@rodrigozhou afaict the lastest release is broken until this is merged

mindaugasrukas (approved):
LGTM

PR #390, opened 2023-05-24
feat: add conditional elasticsearch auth in curl commands

alizdavoodi:
## What was changed
This PR updates the helm chart to conditionally include the Elasticsearch username and password in the `command` for the `check-elasticsearch`, `check-elasticsearch-index`, and `create-elasticsearch-index` containers. 

## Why?
These changes were made to allow the helm chart to be used in situations where the Elasticsearch instance does not require authentication. For example Opensearch is in a private network and doesn't have fine-grained auth in place. (If we pass any arbitrary/default username and password we will get 403)

## Checklist
1. How was this tested:
test these changes by deploying the helm chart to a Kubernetes cluster with an Elasticsearch instance that does not require authentication, to ensure that the commands run successfully without the `--user` flag. Additionally, tested by deploying to a cluster with an Elasticsearch instance that does require authentication, to verify that the `--user` flag is properly included when needed.

### DIFF
--- a/templates/server-deployment.yaml
+++ b/templates/server-deployment.yaml
@@ -74,7 +74,7 @@ spec:
         - name: check-elasticsearch-index
           image: "{{ $.Values.admintools.image.repository }}:{{ $.Values.admintools.image.tag }}"
           imagePullPolicy: {{ $.Values.admintools.image.pullPolicy }}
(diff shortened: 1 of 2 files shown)


mindaugasrukas (approved):
LGTM

PR #394, opened 2023-06-02
FIX remove web ui config in values.yaml

### LINKED ISSUE #393: [Bug] Web UI now configure with env variables instead of configmap
To configure web ui, as shown in the [document](https://github.com/temporalio/helm-charts#updating-temporal-web-config), we use `web.config` in the helm `values.yaml`, the config will be mount as a `configmap` to `temporalio/ui` pod.

Since web ui v2, the mounted configmap volume is removed, as in #300.

In web ui v2, configuration is provided with env variables, but the `values.yaml` and the documentation didn't reflect that.

washanhanzi:
## What was changed

1. remove web ui config in values.yaml
2. remove web config volume in web-deployment.yaml
3. remove web-config config map
4. remove line 280 since the install bash didn't configure web ui auth
5. update the document for web ui configuration with env variable

## Why?
#393

## Checklist

1. Closes #393

2. How was this tested:
manually, 100% craft by hand. since the repo didn't provide any testing tool.

3. Any docs updates needed?
already updated with the PR.


samm-git:
+1, spent ~hour trying to understand why OID does not work when finally found that config map is not mounted to any volume/file.

joshbranham:
Can we get this merged? Also just spent a ton of time on auth that wasn't necessary.

washanhanzi:
@tsurdilo

washanhanzi:
@robholland

robholland (changes requested):
Thanks, looks good, just a few suggestions.

washanhanzi:
All done. By the way, is there any test I can run? The PR has been open for a really long time, and I haven't touched the temporal charts since then. I'm not as confident as I was when I submitted the PR.

robholland:
Nothing automated. We internally use the helm charts to install the base app, but we don't modify the UI so wouldn't pick up on problems. I'll do a manual test before I merge it though.

robholland (approved):

PR #396, opened 2023-06-23
Being able to disable the es schema setup job

SmaineTF1:
## What was changed
Do not run the hook that setups the ES schema if not necessary.

## Why?
Sometimes you don't need to setup the schema if this is done through another process


mindaugasrukas (approved):
LGTM

PR #397, opened 2023-06-23
Being able to set environment variables using 'envFrom'

SmaineTF1:
## What was changed
Add the ability to define environment variables that come from configmaps or secrets

## Why?
Make use of envFrom is pretty handy when, for instance you have all the environment variables defined in Secrets or Configmaps.
That way we only need to have this bit of code if we have all the config env vars in a secret

```yaml
web:
  additionalEnvSecret: 
    - "temporal-web-envvars"
```

### DIFF
--- a/templates/server-deployment.yaml
+++ b/templates/server-deployment.yaml
@@ -121,6 +121,17 @@ spec:
           {{- if or $.Values.server.additionalEnv $serviceValues.additionalEnv }}
           {{- toYaml (default $.Values.server.additionalEnv $serviceValues.additionalEnv) | nindent 12 }}
           {{- end }}
+          {{- if or $serviceValues.additionalEnvCM $serviceValues.additionalEnvSecret }}
+          envFrom:
+            {{- if $serviceValues.additionalEnvCM }}
+            - configMapRef:
+                name: {{ $serviceValues.additionalEnvCM }}
+            {{- end }}
+            {{- if $serviceValues.additionalEnvSecret }}
+            - secretRef:
+                name: {{ $serviceValues.additionalEnvSecret }}
+            {{- end }}
+          {{- end }}
           ports:
             - name: rpc
               containerPort: {{ include (printf "temporal.%s.grpcPort" $service) $ }}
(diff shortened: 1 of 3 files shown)


mindaugasrukas (approved):
LGTM

underrun (approved):

templates/server-deployment.yaml
```diff
@@ -121,6 +121,17 @@ spec:
           {{- if or $.Values.server.additionalEnv $serviceValues.additionalEnv }}
           {{- toYaml (default $.Values.server.additionalEnv $serviceValues.additionalEnv) | nindent 12 }}
           {{- end }}
+          {{- if or $serviceValues.additionalEnvCM $serviceValues.additionalEnvSecret }}
+          envFrom:
+            {{- if $serviceValues.additionalEnvCM }}
```
XaF, line 126: Any reason not to handle those as lists? (vs. scalar)

SmaineTF1: I thought that having a unique secret per component would be enough. I can work on handling a list of CM/Secrets.

SmaineTF1: I just updated in order to take lists? @XaF

XaF: Thanks!

XaF (approved):

SmaineTF1:
What about this PR? Do you need anything else?

sgrzemski:
+1 to that. 
@SmaineTF1 could you please fix the semgrep, so that @XaF @underrun or @mindaugasrukas can merge it?

SmaineTF1:
@sgrzemski sorry, I'm confused but what should I do for that check? This is not obvious to me :/

sgrzemski:
@SmaineTF1 :
image
Sorry, I thought it was in a failed state, but it's just queued to be triggerred manually.

values.yaml
```diff
@@ -84,6 +84,8 @@ server:
   additionalVolumes: []
   additionalVolumeMounts: []
   additionalEnv: []
+  additionalEnvCM: []
```
robholland, line 87: Please use `additionalEnvConfigMap` rather than using a contraction.

robholland (changes requested):
Thanks for the submission, happy to work with you on getting this merged. Please can you rebase this against master.

robholland:
I'm going to close this PR due to inactivity. Please re-open if you'd like to continue the process.

PR #399, opened 2023-07-03
Being able to disable prom annotations

SmaineTF1:
## What was changed
Default behavior should be to have the annotations disabled at the service level.
Note that it is already enabled at the global level by default:

```yaml
server:
  metrics:
    annotations:
      enabled: true
```

## Why?
That makes easier to configure if you use the prometheus operator for instance


SmaineTF1:
Do you need more info?

robholland:
This is backwards incompatible, you are changing the defaults such that new users would not get the annotations. You aren't considering the top-level setting.

PR #400, opened 2023-07-03
Update Temporal version to v1.21.1

mindaugasrukas:
## What was changed
Updated default Temporal version to 1.21.1

## Why?
Use the latest release

## Checklist

1. Closes 
No

2. How was this tested:
CICD

3. Any docs updates needed?
No

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -289,7 +289,7 @@ web:
 
   image:
     repository: temporalio/ui
-    tag: 2.16.1
+    tag: 2.16.2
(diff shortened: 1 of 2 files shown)


values.yaml
```diff
@@ -289,7 +289,7 @@ web:
 
   image:
     repository: temporalio/ui
-    tag: 2.15.0
+    tag: 2.16.1
```
feedmeapples, line 292:
```suggestion
    tag: 2.16.2
```

PR #409, opened 2023-07-14
Adds a PodDisruptionBudget and topologySpreadConstraints to web

sringel:
## What was changed

- added configurable podDisruptionBudget to web component
- added configurable topologySpreadConstraints to web component
- bugfix: topologySpreadConstraints should be of type list instead of map

## Why?

To enable the web component to have a PodDisruptionBudget created and topologySpreadContraints configured due to the requirement of these configurations being enforced by policy.

## Checklist

1. Closes 

2. How was this tested:

`helm template .`

3. Any docs updates needed?

### DIFF
--- a/charts/temporal/templates/web-deployment.yaml
+++ b/charts/temporal/templates/web-deployment.yaml
@@ -85,4 +85,8 @@ spec:
       tolerations:
       {{- toYaml . | nindent 8 }}
     {{- end }}
+    {{- with .Values.web.topologySpreadConstraints }}
+      topologySpreadConstraints:
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
   {{- end }}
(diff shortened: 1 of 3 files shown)


templates/web-pdb.yaml
```diff
@@ -0,0 +1,23 @@
+{{- if $.Values.web.enabled }}
+{{- if and (gt ($.Values.web.replicaCount | int) 1) ($.Values.web.podDisruptionBudget) }}
```
robholland, line 2: I would not filter based on replica count. There is a good chance someone might scale outside of helm later.

robholland (changes requested):

templates/web-pdb.yaml
sringel: This matches the other pdb template
https://github.com/temporalio/helm-charts/blob/master/charts/temporal/templates/server-pdb.yaml#L4

robholland: I'll fix that one in a later PR.

sringel: removed replicaCount check

robholland (approved):

PR #413, opened 2023-07-28
Add Semgrep static analysis

jackdawm:
## What was changed
Add Github workflow to scan PRs with Semgrep.

## Why?
This will eventually be an org-wide required workflow, but Github won't let us have access to it yet, so this PR manually adds scanning.

## Checklist
How was this tested:

Added manually in this way to many other repos with no issue.


robholland:
This is now done automatically.

PR #414, opened 2023-07-31
Add enchanced faultInjection config

alexshtin:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/templates/server-configmap.yaml
+++ b/templates/server-configmap.yaml
@@ -39,11 +39,17 @@ data:
             {{- toYaml . | nindent 12 }}
             {{- end }}
           {{- end }}
-          {{- if $.Values.server.config.persistence.faultinjection}}
-          {{- if $.Values.server.config.persistence.faultinjection.rate }}
+          {{- if or $.Values.server.config.persistence.default.faultInjection $.Values.server.config.persistence.faultInjection}}
           faultInjection:
-            rate: {{ $.Values.server.config.persistence.faultinjection.rate }}
-          {{- end }}
+            {{- if $.Values.server.config.persistence.faultInjection}}
+            {{- if $.Values.server.config.persistence.faultInjection.rate }}
+            rate: {{ $.Values.server.config.persistence.faultInjection.rate }}
+            {{- end }}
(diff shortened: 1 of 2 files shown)


templates/server-configmap.yaml
```diff
@@ -39,11 +39,17 @@ data:
             {{- toYaml . | nindent 12 }}
             {{- end }}
           {{- end }}
-          {{- if $.Values.server.config.persistence.faultinjection}}
-          {{- if $.Values.server.config.persistence.faultinjection.rate }}
+          {{- if or $.Values.server.config.persistence.default.faultInjection $.Values.server.config.persistence.faultInjection}}
```
alexshtin, line 42: Old simple "just `rate`" format is still supported to keep backward compatibility.

values/values.cassandra.yaml
```diff
@@ -8,6 +8,24 @@ server:
       default:
         driver: "cassandra"
 
+#        faultInjection:
```
alexshtin, line 11: I leave it here for future references as example on how to configure fault injection.

templates/server-configmap.yaml
mindaugasrukas: We may not need this backward compatibility.
I'd go with a simple version without backward compatibility.

templates/server-configmap.yaml
```diff
@@ -39,11 +39,17 @@ data:
...
           faultInjection:
-            rate: {{ $.Values.server.config.persistence.faultinjection.rate }}
-          {{- end }}
+            {{- if $.Values.server.config.persistence.faultInjection}}
+            {{- if $.Values.server.config.persistence.faultInjection.rate }}
+            rate: {{ $.Values.server.config.persistence.faultInjection.rate }}
+            {{- end }}
+            {{- else if $.Values.server.config.persistence.default.faultInjection}}
```
mindaugasrukas, line 48: `if` is not necessary for `with`: https://helm.sh/docs/chart_template_guide/control_structures/#modifying-scope-using-with

mindaugasrukas (approved):

PR #415, opened 2023-08-03
fix: sidecarContainers should be an array, not a dict

kuzmik:
## What was changed

Very simple change: I changed the datatype of `sidecarContainers` to `[]` (from `{}`).

## Why?

When running Tanka, we've been getting this warning:

`coalesce.go:220: warning: cannot overwrite table with non table for temporal.server.sidecarContainers (map[])`

It was rendering the YAML fine, but when CI runs people kept thinking there was an actual issue.

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -49,7 +49,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.26.0
+version: 0.26.1
 
 # This is the version number of the application being deployed. This version number should be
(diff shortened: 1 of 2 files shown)


robholland:
Thanks for your submission. This fixed has been merged, unfortunately I wasn't able to get it to merge via the PR for some reason.

kuzmik:
> Thanks for your submission. This fixed has been merged, unfortunately I wasn't able to get it to merge via the PR for some reason.

that's ok with me, as long as it gets fixed. it's not like this was a super complicated PR :D

PR #424, opened 2023-09-18
feat: updated grafana and prometheus helm dependencies

dotdc:
## What was changed

Updated Temporal's Helm chart dependencies:
- Grafana from `5.0.10` to `8.0.2`.
- Prometheus from `15.1.3` to `25.22.0`.

While it's a huge bump, I've raised the temporal minor version as it shouldn't be a breaking change in most cases.
Let me know if you prefer to bump the chart's version differently.

## Why?

Take advantage of products updates while reducing security risks. 

## Checklist

1. Closes 

N/A

2. How was this tested:

```bash
helm template temporal . -n temporal | kubectl apply -f -
```

3. Any docs updates needed?

N/A

### DIFF
--- a/charts/temporal/Chart.yaml
+++ b/charts/temporal/Chart.yaml
@@ -39,7 +39,7 @@ dependencies:
 type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.44.0
+version: 0.43.0
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
 appVersion: 1.24.2
--- a/charts/temporal/Chart.yaml
+++ b/charts/temporal/Chart.yaml
@@ -18,7 +18,7 @@ dependencies:
     condition: cassandra.enabled
   - name: prometheus
     repository: https://prometheus-community.github.io/helm-charts
-    version: 15.1.3
+    version: 25.22.0
     condition: prometheus.enabled
   - name: elasticsearch
     repository: https://helm.elastic.co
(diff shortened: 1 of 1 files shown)


dotdc:
Hi,

Could someone review this PR?
I can also bump the charts further as new versions came out since then.

cc @mindaugasrukas

mindaugasrukas:
@alexshtin took over ownership of this project.
While this PR LGTM, In general, the architecture of having a direct dependency on Grafana, Prometheus, Cassandra, and Elasticsearch is incorrect. That said, you can install whatever indirect dependencies you like and don't rely on this helm-chart. For example, if you want to switch from Cassandra to Postgres, you don't need to add this direct dependency here, and you can install whatever Postgres version you like. The only requirement here is to have the correct dependencies installed.

dotdc:
I was not sure for ownership, but thank you for your reply 👍
We have a workaround, but it would be nice to update the default for everyone.

robholland:
Please feel free to re-bump these. Leave the chart version as-is though, that gets bumped by our tooling on release. I will ensure it gets reviewed and merged.

dotdc:
Hi @robholland,
versions have been bumped again, also left the chart version as-is like requested.

robholland:
I've tested that end to end metrics from Temporal Server -> Grafana is working. I'll merge this, but please note that we'll soon be removing grafana and prometheus from our chart.

node-exporter will not come up without adjustments on Docker Desktop on Darwin, but ignoring that as we'll soon remove these dependencies anyway (as they aren't actually dependencies at all).

robholland (approved):

PR #429, opened 2023-10-01
feat: added a new conditional for web to allow overriding of the entry point.

flashpaper12:
## What was changed
Added a new conditional for web to allow overriding of the entry point.

## Why?
This is because I would like to run a custom script inside the pod before the service gets started

2. How was this tested:
Changes were tested locally, and chart can be deployed successfully to k8s 

3. Any docs updates needed?


robholland:
Thanks for your contribution, but I don't think this is something we should support. If you need to do something custom, you can use a custom image.

PR #431, opened 2023-10-13
Add support of pod annotations for service job

jeenadeepak:
## What was changed

Add support for pod annotations

Istio run with each pod due to which one time job/pod are not successfully terminal, istio is not required here as db schema setup and update are one time job, hence skipping istio by adding addition skip parameter.

## Checklist

1. Closes 
https://github.com/temporalio/helm-charts/issues/353
2. How was this tested:

I have deployed this on eks with istio enabled namespace, and it is working fine.

3. Any docs updates needed?

### DIFF
--- a/templates/server-job.yaml
+++ b/templates/server-job.yaml
@@ -35,6 +35,10 @@ spec:
         app.kubernetes.io/version: {{ .Chart.AppVersion | replace "+" "_" }}
         app.kubernetes.io/component: database
         app.kubernetes.io/part-of: {{ .Chart.Name }}
+      annotations:
+        {{- with .Values.serviceJob.podAnnotations }}
+          {{- toYaml . | nindent 8 }}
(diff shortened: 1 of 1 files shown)


templates/server-job.yaml
```diff
@@ -35,6 +35,10 @@ spec:
         app.kubernetes.io/version: {{ .Chart.AppVersion | replace "+" "_" }}
         app.kubernetes.io/component: database
         app.kubernetes.io/part-of: {{ .Chart.Name }}
+      annotations:
+        {{- with .Values.serviceJob.podAnnotations }}
```
debugger24, line 39: What is `serviceJob` here ?

jeenadeepak: "serviceJob" variable refer to properties of "templates/server-job.yaml" file.

PR #436, opened 2023-10-24
fix: Use visibility server when defining visibility config

bakjos:
## Why?

When the default store uses `cassandra` and the visibility store `sql` the config was not including the `sql` part

## Checklist

1. Closes 

2. How was this tested:

Using the following values 

```yaml
server:
  replicaCount: 1
  config:
    persistence:
      default:
        cassandra:
          hosts:
            - cassandra
          port: 9042
      visibility:
        driver: "sql"
        sql:
          driver: "postgres"
          host: "postgres"
          port: 5432
          database: "temporal_visibility"
          user: "postgres"
          password: "postgres"
          maxConns: 20
          maxConnLifetime: "1h"
```

3. Any docs updates needed?


robholland (approved):

robholland:
Thanks :)

PR #450, opened 2024-01-04
Add additional Init Containers and Volume Mounts

askoriy:
## What was changed
Add the option co configure additional init containers and volume mounts

## Why?
In our case to get certificates from Vault and mount as files

### DIFF
--- a/templates/admintools-deployment.yaml
+++ b/templates/admintools-deployment.yaml
@@ -37,6 +37,10 @@ spec:
       {{- end }}
     spec:
       {{ include "temporal.serviceAccount" . }}
+      initContainers:
+        {{- if $.Values.admintools.additionalInitContainers }}
+        {{- toYaml $.Values.admintools.additionalInitContainers | nindent 8}}
+        {{- end }}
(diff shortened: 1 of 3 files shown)


templates/admintools-deployment.yaml
```diff
@@ -77,6 +85,10 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      volumes:
```
robholland, line 88: Please move this inside the conditional.

templates/admintools-deployment.yaml
```diff
@@ -37,6 +37,10 @@ spec:
       {{- end }}
     spec:
       {{ include "temporal.serviceAccount" . }}
+      initContainers:
```
robholland, line 40: Please move this inside the conditional.

robholland (changes requested):

enricojonas:
@robholland  @askoriy  can we get this merged? We need this functionality.

PR #451, opened 2024-01-04
Fix rendering for elasticsearch visibility storage

askoriy:
## What was changed
To specify Elasticsearch as only visibility storage, the value `server.config.persistence.visibility.driver` should be set to `elasticsearch`.
This omits configuring Cassandra as visibility storage coming from default values.

## Why?
When Cassandra was disabled and external Postgres was used as default storage and external Elasticsearch for visibility storage, helm failed to install because created inconsistent resources and configuration.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/templates/server-deployment.yaml
+++ b/templates/server-deployment.yaml
@@ -109,11 +109,13 @@ spec:
                 secretKeyRef:
                   name: {{ include "temporal.persistence.secretName" (list $ "default") }}
                   key: {{ include "temporal.persistence.secretKey" (list $ "default") }}
+          {{- if ne (include "temporal.persistence.driver" (list $ "visibility")) "elasticsearch" }}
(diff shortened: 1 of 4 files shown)


templates/server-deployment.yaml
```diff
@@ -109,11 +109,13 @@ spec:
                 secretKeyRef:
                   name: {{ include "temporal.persistence.secretName" (list $ "default") }}
                   key: {{ include "temporal.persistence.secretKey" (list $ "default") }}
+          {{- if ne (include "temporal.persistence.driver" (list $ "visibility")) "elasticsearch" }}
```
rodrigozhou, line 112: You could use the existing `$.Values.elasticsearch.enabled` instead, no?

rodrigozhou (changes requested):

robholland:
Fixed by #522

PR #452, opened 2024-01-08
Ability to specify a loadBalancerIP for the frontend service

richiejenkins:
## What was changed
Adds the ability to specify a loadBalancerIP for the frontend service.

## Why?
Allow static IP for load balancer.

## Checklist

1. Closes: None 

2. How was this tested: `helm template`


robholland:
Thank you for your submission, but the loadBalancerIP field is deprecated, so we won't be merging this.

PR #411, opened 2023-07-20
Configure SQL TLS environment variables in server-job

grzegorz8:
## What was changed

Added support for `SQL_TLS*` properties in schema jobs (`server-job.yaml`).

## Why?

Currently, when PostgreSQL database is secured with TLS, schema jobs are not properly configured.

In the Pull Request we reuse `server.config.persistence.*.sql.tls`, `server.config.additionalVolumes` and `server.config.additionalVolumeMounts` properties..

## Checklist

1. Closes 
No related issues.

2. How was this tested:

3. Any docs updates needed?

No updates needed.

PR #456, opened 2024-01-26
Set SQL-TLS when sql.tls.enabled is true

alexandermarston:
## What was changed
This change adds `SQL_TLS` as an environment variable to the schema jobs, when you have TLS enabled for your SQL Driver.

## Why?
The Schema Jobs are unable to run against a PostgreSQL server that requires TLS, as it does not currently respect your TLS settings.

## Checklist
1. Closes
None

3. How was this tested:
Tested on a 1.28 Kubernetes cluster, with an AWS RDS Proxy being used as the PSQL host.

4. Any docs updates needed?
N/A

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -49,7 +49,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.32.0
+version: 0.33.0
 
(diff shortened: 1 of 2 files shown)


syugoman-armenotech:
it looks like this is a similar PR - https://github.com/temporalio/helm-charts/pull/411

alexandermarston:
closing in favour of #411

Issue #330, opened 2022-11-03
[Feature Request] Tag all releases


gregbrowndev:
### Is your feature request related to a problem? Please describe.

Hi, I noticed that this repo has not created GitHub tagged releases for all the changes to the helm chart to track the underlying releases of the Temporal server. E.g. the [latest commit bumps to v1.18.4](https://github.com/temporalio/helm-charts/commit/fd5d50f84d18eb2a2c3b7d8f7f8f3b2cd135d991) but this does not appear as a GitHub release. Do you intent to only create releases for major versions or is this just a small oversight?

A GitHub release would be preferable as we currently manage packaging the Helm charts in this repository ourselves. With a tag, we can detect new releases, checkout the specific commit, get the version number from the tag, package the chart archive and push to our helm repository.

### Describe the solution you'd like

Tag each commit that bumps the Chart's `appVersion` and image tag of the underlying Temporal server so that it appears as a GitHub release.

Thank you

Issue #45, opened 2020-07-14
Publish temporal helm chart to a helm chart repository (e. g. Helm Hub) 


markmark206:
Please publish Temporal Helm chart to a helm chart repository (such as [Helm Hub](https://hub.helm.sh/charts)), so it is easier to install it.

[Brought to our attention](https://community.temporal.io/t/is-the-helm-chart-published-to-a-helm-repository-somewhere/182?u=markmark) by Kyle, thank you!

sagikazarmark:
Helm hub is not actually a chart repository, and it's being replaced by [Artifact Hub](https://artifacthub.io/)

For simplicity, I'd consider making this repository a helm chart repository using the chart releaser action: https://github.com/helm/chart-releaser-action

lerminou:
Hi @sagikazarmark , any news for this github as chart repo  ?

loopingrage:
Maybe poke the Bitnami folks to take it on?

joebowbeer:
This would be greatly appreciated, and should be easy to resolve.

The simplest approach is to use the `chart-releaser-action`:

https://github.com/helm/chart-releaser-action

This requires a `gh-pages` branch to hold the published charts.

After publishing, the temporalio helm repo can be added via:

`helm repo add temporalio https://temporalio.github.io/helm-charts`

LocalStack publishes their chart here:

https://github.com/localstack/helm-charts/blob/main/.github/workflows/chart-release.yaml

Their helm repo is: https://localstack.github.io/helm-charts

sagikazarmark:
Some other examples:

- https://github.com/sagikazarmark/helm-charts
- https://github.com/dexidp/helm-charts

joebowbeer:
I forked and published from https://github.com/joebowbeer/temporalio-helm-charts

loopingrage:
Ay update here? Seems like a simple thing to do that would save the community a bunch of time and confusion.

abatilo:
Bump. Hard to seriously consider this project until there's an easy way to install this chart

hazcod:
Pinging.

ErlanBelekov:
Any updates on this?

hazcod:
This should be an easy fix so really looking to a quick fix here.

franklinkim:
Some references for anyone else coming along this path:

- #45
- #147 
- #229
- #189
- #42

Looking forward seeing to this progress

IkePCampbell:
Any update on this?

[...]

Issue #229, opened 2021-09-28
Publish helm charts repo via GitHub pages


andrey-dubnik:
Hi, I was looking at the option to consume helm in a bit more convenient way which is possible via the GitHub pages and Helm releaser action. In order to implement that I've refactored the current helm repo in a way which would make a deployment flexible with possibility to add more charts on a later stage if needs to.

Repo reference: https://github.com/andrey-dubnik/helm-charts
Helm landing page: https://andrey-dubnik.github.io/helm-charts/

This way you can add helm as `helm repo add temporalio https://temporalio.github.io/helm-charts`

I can make a PR if this is Ok

underrun:
this is definitely in plan

joebowbeer:
For background and previous PR:

* https://github.com/temporalio/helm-charts/issues/45
* https://github.com/temporalio/helm-charts/pull/147

IkePCampbell:
Any news on this?

pzduniak:
This still makes it incredibly annoying to deploy Temporal in systems that rely on Helm repos.

Please dedicate the couple hours of effort to set it up. I did it for https://github.com/mailgun/gubernator.

froblesmartin:
A similar issue exists, it would be better to close one so that there is a single place to comment on the Helm chart publication
https://github.com/temporalio/helm-charts/issues/45

robholland:
We now publish to https://go.temporal.io/helm-charts

monotek:
OCI images would be nice too :)

Issue #445, opened 2023-11-22
[Bug] ElasticSearch index not created when using existing SQL/Cassandra


RonaldGalea:
### What are you really trying to do?

Deploy Temporal with an existing Database.

### Describe the bug

When an exiting DB is used, the suggested configuration is disabling the schema setup and update:

```yaml
schema:
  setup:
    enabled: false
  update:
    enabled: false
```

However, this causes the the index creation job not to run either: https://github.com/temporalio/helm-charts/blob/master/templates/server-job.yaml#L347

There should likely be a separate flag controlling the ElasticSearch index creation.

### Minimal Reproduction

Just run any of the "Install with your own MySQL/PostgreSQL/Cassandra" examples. All server services will be stuck in Init "waiting for elasticsearch index to become ready"

### Additional context

There is [this](https://community.temporal.io/t/temporal-pods-not-coming-up-with-index-not-found-exception-error/9010) post on the community forum which might be related.

DanielCalvo:
I stumbled upon this today while trying to configure temporal with an external PostgreSQL DB. I also found a workaround, but it isn't pretty.

We're deploying this through an ArgoCD application, and this is what it looks like: 

```
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
    name: temporal
    namespace: argocd
spec:
    syncPolicy:
        automated:
            selfHeal: true
            prune: true
    project: default
    destination:
        server: https://kubernetes.default.svc
        namespace: temporal
    source:
        path: charts/temporal
        repoURL: https://github.com/temporalio/helm-charts
        targetRevision: temporal-0.33.0
        helm:
            releaseName: temporal
            values: |-
                replicaCount: 1
                postgresql:
                  enabled: true
                prometheus:
                  enabled: true
                elasticsearch:
                  enabled: true
                grafana:
                  enabled: true
                cassandra:
                  enabled: false
                schema:
                  setup:
                    enabled: true
                  update:
                    enabled: true
                server:
                  config:
                    persistence:
                      default:
                        driver: sql
                        sql:
                          driver: postgres
                          host: temporal(...).eu-west-1.rds.amazonaws.com
                          port: 5432
                          user: postgresql
                          password: xxxx
                      visibility:
                        driver: sql
                        sql:
                          driver: postgres
                          host: temporal(...).eu-west-1.rds.amazonaws.com
                          port: 5432
                          user: postgresql
                          password: xxxx
```
[...]

PR #457, opened 2024-01-30
Release and publish helm charts via a Github Action Release workflow.

robholland:
## What was changed

Added a new workflow that automates the release of the Helm chart. This workflow also makes the Helm chart releases available via a helm repo: https://go.temporal.io/helm-charts.

Installation via the helm repo is not yet documented, a follow-up PR will add the repo to the documentation.

## Why?

This automates some of the manual work currently performed by OSS team. This also allow users to install Temporal via Helm without having to download the repo, although they will probably still need to refer to the (rather long) values files to achieve their desired configuration.

Fixes: https://github.com/temporalio/helm-charts/issues/330

Also technically fixes: https://github.com/temporalio/helm-charts/issues/45 https://github.com/temporalio/helm-charts/issues/229, but documentation needs to be updated to show users how to install from a repo.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?
Docs will need updating to highlight the new installation method.


.github/workflows/publish-charts.yml
```diff
@@ -6,13 +6,71 @@ on:
...
+          else
+            let PATCH+=1
+          fi
+          NEW_VERSION="$MAJOR.$MINOR.$PATCH"
+
+          sed -i -e "s/^version: .*/version: $NEW_VERSION/" Chart.yaml
+          sed -i -e "s/^appVersion: .*/appVersion: ${{ github.event.inputs.version }}/" Chart.yaml
+
+          git commit -m "Update Chart to $NEW_VERSION, Temporal v${{ github.event.inputs.version }}" Chart.yaml
+          git push origin
```
dandavison, line 68: This is really nice; I learned a couple of things from it; `grep -m` and I didn't know about the `let` arithmetic at all, I thought one had to use `((...))`. Could use `sed -E` to make the regex more readable?)

dandavison (approved):
LGTM!

fakeNews-jpg:
@robholland , for information, when i try https://go.temporal.io/helm-charts i have the following error when trying to install temporal.io : "gzip: invalid header"
To reproduce: get a rancher, add this repo, go to app catalog, install the app

Do you have an idea why it does not work? 

Thanks a lot again for your amazing work :)

fakeNews-jpg:
I just tried again when i saw the new version 0.33, and this time it seems to work at first sight ! thanks a lot @robholland  :)

fakeNews-jpg:
I just have the issue https://github.com/temporalio/helm-charts/issues/445 popping when trying to use an existing postgresql on my helm

wrbbz:
Are there any plans to add a folow-up documentation with Helm repo usage?

robholland:
Yep, it's on my list: https://github.com/temporalio/helm-charts/issues/458

PR #463, opened 2024-02-07
changes for adding volumes to admin tools

dhurvas:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


charts/temporal/templates/admintools-deployment.yaml
```diff
@@ -37,6 +37,10 @@ spec:
       {{- end }}
     spec:
       {{ include "temporal.serviceAccount" . }}
+      volumes:
```
robholland, line 40: Please move this inside the conditional.

robholland (changes requested):

robholland:
Fixed via #450.

PR #459, opened 2024-02-02
Remove Cassandra visibility support

rodrigozhou:
## What was changed

Remove Cassandra visibility support

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

PR #467, opened 2024-02-19
Fix condition for rendering visibility config

### LINKED ISSUE #466: Visibility config doesn't get rendered if visibility and default store types are different

I'm trying to install Temporal with Cassandra as the default store and PostgreSQL as the visibility store. 

### Describe the bug

While rendering visibility store config, instead of checking for .visibility key, it was checking for .default. (The chart works when both your default and visibility store are same)

Responsible line of code: https://github.com/temporalio/helm-charts/blob/master/charts/temporal/templates/server-configmap.yaml#L60

### Minimal Reproduction

```yaml
server:
  config:
    persistence:
      default:
        cassandra:
         # cassandra config
        driver: cassandra
      defaultStore: default
      visibility:
        driver: sql
        sql:
           # sql config
          driver: postgres12
```

### Environment/Versions

Not applicable.

np-we360:
While rendering visibility store config, instead of checking for .visibility key, it was checking for .default. Works when both your default and visibility store are same. Otherwise, doesn't.

## What was changed
The condition for rendering visibility config in the server configmap was fixed.

## Why?
This was a typo-like bug that was fixed.

## Checklist

1. Closes #466 .

2. How was this tested:
By rendering using helm template.

3. Any docs updates needed?
No.

### DIFF
--- a/charts/temporal/templates/server-configmap.yaml
+++ b/charts/temporal/templates/server-configmap.yaml
@@ -57,7 +57,7 @@ data:
             {{- end }}
           {{- end }}
         visibility:
-          {{- if eq (include "temporal.persistence.driver" (list $ "default")) "sql" }}
+          {{- if eq (include "temporal.persistence.driver" (list $ "visibility")) "sql" }}
           sql:
             pluginName: "{{ include "temporal.persistence.sql.driver" (list $ "visibility") }}"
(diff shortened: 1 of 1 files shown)


RajAnthari:
Maybe #459 missed to remove Cassandra visibility changes from `values.yaml` as well.

image

I'm not sure whether it can be fixed with this PR or requires a new PR.

PR #480, opened 2024-04-09
Allow to skip database creation

### LINKED ISSUE #440: [Feature Request] Allow to skip database creation

I want to use an existing database with an user that don't have the permission to create the database.

When running the helm chart I get the following error:
```
Unable to create SQL database.	{"error": "pq: permission denied to create database", "logging-call-at": "handler.go:94"}
```

It would be good to have a flag in the helm chart to switch off the db creation and keep the schema creation active (so setting schema.setup.enabled=false doesnt work for me)

VLZZZ:
## What was changed

This PR adds an ability to skip database creation. The default values ensures that the change will be backwards compatible with an existing installations.
## Why?

Currently it's not possible to use already existing database when deploying Temporal using this Helm chart
## Checklist

1. Closes #440

2. How was this tested:

I've performed a fresh install with:
```
schema:
  createDatabase:
    enabled: false
```

3. Any docs updates needed?

### DIFF
--- a/charts/temporal/templates/server-job.yaml
+++ b/charts/temporal/templates/server-job.yaml
@@ -49,6 +49,7 @@ spec:
           imagePullPolicy: {{ .Values.cassandra.image.pullPolicy }}
           command: ['sh', '-c', 'until cqlsh {{ include "cassandra.host" $ }} {{ .Values.cassandra.config.ports.cql }} -e "SHOW VERSION"; do echo waiting for cassandra to start; sleep 1; done;']
         {{- end }}
+        {{- if .Values.schema.createDatabase.enabled }}
         {{- range $store := (list "default" "visibility") }}
         {{- $storeConfig := index $.Values.server.config.persistence $store }}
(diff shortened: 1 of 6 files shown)


zaenk:
I just wanted to do this exact same thing! Thanks for opening this PR!

charts/temporal/Chart.yaml
```diff
@@ -49,7 +49,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.36.0
+version: 0.37.0
```
robholland, line 52: Please revert this change, the version is bumped automatically by our tooling.

robholland (changes requested):

robholland (approved):

robholland:
Thanks for your contribution :) For future reference: if you make changes to address PR feedback, if you re-request feedback from the relevant person, it makes it more obvious the PR is ready for review again.

PR #488, opened 2024-05-13
Ensure appVersion is used by default as the server image tag.

ktulsyan:
## What was changed
Updated image versions for frontend/matching/worker/history/admintools to 1.23.1

## Why?
Chart.yaml and values.yaml were out of sync

## Checklist

2. How was this tested:
Not tested 

3. Any docs updates needed?

### DIFF
--- a/charts/temporal/Chart.yaml
+++ b/charts/temporal/Chart.yaml
@@ -49,7 +49,7 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 0.37.0
+version: 0.38.0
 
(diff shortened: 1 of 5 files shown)


robholland:
Thank you for contributing :)

Our systems bump the appVersion in the Chart.yaml but we missed that our templates didn't use that value. Please use templating in the server templates, and replace the tags in the values file with empty ones.

Some more context/suggestions here: https://stackoverflow.com/questions/58065225/image-tag-in-values-yaml-vs-appversion-in-chart-yaml.

Please let me know if that's not clear, you can find me on public Slack if that's easiest.

robholland (approved):
Looks great, thanks!

PR #489, opened 2024-05-17
Fix default image tag type in values 

Blamas:
## What was changed

Fix the default type of the image tag from map `{}` to string `""` in the values file.
Issue introduced in https://github.com/temporalio/helm-charts/pull/488

## Why?

It's causing these errors when rendering the chart
```
coalesce.go:286: warning: cannot overwrite table with non table for temporal.temporal.server.image.tag (map[])
``` 

## Checklist

1. Closes 
   None

2. How was this tested:
   By rendering helm templates locally

3. Any docs updates needed?
   No

### DIFF
--- a/charts/temporal/values.yaml
+++ b/charts/temporal/values.yaml
@@ -21,7 +21,7 @@ server:
   sidecarContainers: {}
   image:
     repository: temporalio/server
-    tag: {}
+    tag: ""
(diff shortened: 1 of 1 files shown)


robholland (approved):

robholland:
Thanks :)

Issue #304, opened 2022-07-11
[Bug] Temporal UI not honoring Auth configs


kkcmadhu:
### Web Auth  (outh) config setting breaks

Document says web config can be mounted.  refer https://github.com/temporalio/helm-charts#updating-temporal-web-config 
Recently temporal-web is replaced with tempoal-ui https://github.com/temporalio/helm-charts/commit/098fab57a4c19855b46692aef0df78e2bcf2e336

Looks like temporal-ui is not honoring the  auth configs

### Environment/Versions
1.17

nodnallquick:
Hi I am also having an issue with this, from what I can tell this is because the volumeMount section of the web-deployment.yaml was removed in 1.17.1. Can the docs either be updated to indicated this is no longer supported or the chart updated to mount the volumes again?

Igogrek:
I'm not sure what was the reason for this part to be removed from the chart, but as temp solution ENV vars from here could be used for the ui container https://github.com/temporalio/ui-server/tree/main/docker#quickstart-for-production

i.e. add these to the deployment:
```yaml
    env:
    - name: TEMPORAL_AUTH_ENABLED
      value: "true"
    - name: TEMPORAL_AUTH_PROVIDER_URL
      value: "https://accounts.google.com"
    - name: TEMPORAL_AUTH_CLIENT_ID
      value: "xxxxx-xxxx.apps.googleusercontent.com"
    - name: TEMPORAL_AUTH_CLIENT_SECRET
      value: "xxxxxxxxxxxxxxx"
    - name: TEMPORAL_AUTH_CALLBACK_URL
      value: "https://xxxx.com:8080/auth/sso/callback "
```

EslamNasser:
The issue is still there with the latest helm chart, the config passed in the values files as a volume is not being mounted.

drileyst:
> I'm not sure what was the reason for this part to be removed from the chart, but as temp solution ENV vars from here could be used for the ui container https://github.com/temporalio/ui-server/tree/main/docker#quickstart-for-production
> 
> i.e. add these to the deployment:
> 
> ```yaml
>     env:
>     - name: TEMPORAL_AUTH_ENABLED
>       value: "true"
>     - name: TEMPORAL_AUTH_PROVIDER_URL
>       value: "https://accounts.google.com"
>     - name: TEMPORAL_AUTH_CLIENT_ID
>       value: "xxxxx-xxxx.apps.googleusercontent.com"
[...]

PR #490, opened 2024-05-20
Add auth and database features

dcaputo-harmoni:
## What was changed

This PR includes two primary functionality enhancements to the helm chart:

- Addition of Auth settings to both server and web
- Allowing database username to optionally be specified in a secret alongside the password

## Why?

I'll address each area separately:

**Auth** - production environments need some form of authentication.  To use authentication, it was required to be enabled in both the server and web components.  Additionally, an `internal-frontend` service was added (per the v1.20.0 release) to allow the workflow service to function.  This also addresses an open issue (#304).

**Database Creds** - many dynamic database creds generation systems (such as hashicorp vault) generate both the username and password dynamically, which is more secure than just rotating the password.  The functionality has been updated to allow either or both of these to be optionally set via a secret with any or both of `username` and `password` keys.

## Checklist

1. Closes  #304

2. How was this tested:

Tested a number of times using various alternatives of the configurations.

3. Any docs updates needed?

Included documentation via commented out items in values.yaml

### DIFF
--- a/charts/temporal/templates/_helpers.tpl
+++ b/charts/temporal/templates/_helpers.tpl
@@ -85,6 +85,22 @@ Source: https://stackoverflow.com/a/52024583/3027614
 {{- end -}}
 {{- end -}}
 
+{{- define "temporal.internal-frontend.grpcPort" -}}
+{{- if index $.Values.server "internal-frontend" "service" "port" -}}
+{{- index $.Values.server "internal-frontend" "service" "port" -}}
+{{- else -}}
+{{- 7236 -}}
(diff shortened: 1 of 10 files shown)


robholland:
Thank you for your contribution :)

Please don't mix concerns in the same PR, split out: 1. database credentials 2. internal-frontend support

For the auth config side, I'm not sure about proxying all of the values this way. Web can be configured via additionalEnv and for server auth I think maybe just have a `auth: { enabled: true, config: { ... } }` that gets copied as well.

PR #492, opened 2024-06-05
Use tplvalues.render for templating inside values

vogre:
## What was changed
Use templating for `web.service` annotations in the same way as `server.service` is defined https://github.com/temporalio/helm-charts/blob/bad8c74311717f2121579f64e2e6b1670fc19818/charts/temporal/templates/server-service.yaml#L15.

## Why?
A very cool trick to allow passing templated values inside the values.yaml file instead of relying on an outside templating engine

## Checklist

1. Closes 

2. How was this tested:
Added a template in `values.yaml` and ran `helm template`.

3. Any docs updates needed?
No

### DIFF
--- a/charts/temporal/templates/web-service.yaml
+++ b/charts/temporal/templates/web-service.yaml
@@ -11,10 +11,9 @@ metadata:
     app.kubernetes.io/version: {{ .Chart.AppVersion | replace "+" "_" }}
     app.kubernetes.io/component: web
     app.kubernetes.io/part-of: {{ .Chart.Name }}
-{{- with .Values.web.service.annotations }}
-  annotations:
-{{ toYaml . | indent 4 }}
-{{- end }}
+  {{- if .Values.web.service.annotations }}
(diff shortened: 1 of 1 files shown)


RoiAuthomize (approved):

robholland:
I know we already do this, and I'm happy to merge, but could you please explain what the use case is here so I can learn?

vogre:
@robholland - we use it roughly like this:

e.g. by `helm template --set foo=bar -f values_web.yaml`, 
and `values_web.yaml` of 
```yaml
web:
  service:
    annotations:
      url: https://{{ $.Values.foo }}-temporal.example.com
```

We would utilize the templating features of helm, inside the `values.yaml` file (usually the templating only works in the `templates/` directory.

Since this behavior is actually surprising - we could also set the value directly by 

```
helm template --set 'web.service.annotations.url=https://bar-temporal.example.com'
```

Although it's a nice feature to allow extensibility of the template.

robholland (approved):

PR #493, opened 2024-06-06
Support new admintools image tag format

robholland:
## What was changed
Amended the admintools image tag to match the current release. Adjusted the release action to prompt for server, ui and admintools versions separately.

## Why?
admintools versioning has changed to no longer follow just the server version.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


alexshtin (approved):
Thanks for fixing this.

PR #485, opened 2024-05-01
Update Cassandra host URLs to remove the ".cluster.local" suffix

cege7480:
## What was changed
No longer hardcoding the cluster.local suffix on various service reference entries.

## Why?
Hardcoding breaks when the cluster name isn't cluster.local.
Relative path will work unless you've done something super weird with your dns.
## Checklist

1. Closes #469 

2. How was this tested:
Ran the helm chart...  installed.  Boom.  
Successfully able to resolve using relative domain names.

3. Any docs updates needed?
Nah.

PR #500, opened 2024-06-14
Revert "Update Cassandra host URLs to remove the ".cluster.local" suffix (#485)"

prathyushpv:
This reverts commit b25c4fcd0e2661d1b18e3bc018ac957d4193448c.

## What was changed
Revert the change to remove the ".cluster.local" suffix.

## Why?
Just noticed that some of our test cells could not come up after this change.
Our busybox container was not able to resolve cassandra pods without .cluster.local


robholland (approved):

robholland:
I'll do a follow-up PR to enable customising "cluster.local" instead of this.

PR #501, opened 2024-06-15
Remove hard coded cluster.local references.

robholland:
Switch check-cassandra-service init container to use nc. nslookup in recent busybox images is broken and doesn't obey resolv.conf, meaning it won't check k8s search domains.

Alternative fix for #485


prathyushpv:
Should we have to pass any config value for it to use cassandra host's full name?

robholland:
No, it should work out of the box.

PR #506, opened 2024-06-18
Fixing condition for creating database

prathyushpv:
## What was changed
Fixing the condition where we create database.

## Why?
database is not created for sql


robholland:
Fixed by #522

PR #511, opened 2024-06-25
Enable HTTP API for Nexus

bergundy:
- Enable HTTP API on the frontend
- Expose it on the frontend service

### DIFF
--- a/charts/temporal/templates/server-configmap.yaml
+++ b/charts/temporal/templates/server-configmap.yaml
@@ -122,6 +122,7 @@ data:
       frontend:
         rpc:
           grpcPort: {{ include "temporal.frontend.grpcPort" $ }}
+          httpPort: {{ include "temporal.frontend.httpPort" $ }}
           membershipPort: {{ include "temporal.frontend.membershipPort" $ }}
           bindOnIP: "0.0.0.0"
 
@@ -160,6 +161,7 @@ data:
           initialFailoverVersion: 1
           rpcName: "temporal-frontend"
           rpcAddress: "127.0.0.1:7233"
+          httpAddress: "127.0.0.1:7243"
     {{- end }}
 
     {{- if $.Values.server.config.dcRedirectionPolicy }}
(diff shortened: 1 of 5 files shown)


charts/temporal/templates/server-configmap.yaml
```diff
@@ -160,6 +161,7 @@ data:
           initialFailoverVersion: 1
           rpcName: "temporal-frontend"
           rpcAddress: "127.0.0.1:7233"
+          httpAddress: "127.0.0.1:7243"
```
pdoerner, line 164: Do we need to worry about / can we generate a UUID for cluster ID here? Now that we are relying on it being present for https://github.com/temporalio/temporal/pull/6184

charts/temporal/templates/server-service.yaml
```diff
@@ -24,6 +24,11 @@ spec:
       {{- if hasKey .Values.server.frontend.service "nodePort" }}
       nodePort: {{ .Values.server.frontend.service.nodePort }}
       {{- end }}
+    - port: {{ .Values.server.frontend.service.httpPort }}
+      targetPort: http
+      protocol: TCP
+      name: http
+      # TODO: Also expose HTTP port
```
pdoerner, line 31: Is this TODO still relevant? If so could you be a little more specific since it looks like this block is for exposing the HTTP port

bergundy, line 31:
```suggestion
      # TODO: Allow customizing the node HTTP port
```

charts/temporal/templates/server-configmap.yaml
bergundy: The cluster ID is auto-generated.

```
$ temporal operator cluster describe --output json
{
  ...
  "clusterId": "855e5c4b-d994-4c72-a09b-3c18e5161bb2",
  ...
}

```

charts/temporal/templates/server-service.yaml
bergundy: Fixed the comment, thanks.

pdoerner (approved):

Issue #523, opened 2024-07-03
[Bug] Postgres/MySQL databases are not provisioned on fresh install


ogusak:
### What are you really trying to do?

Install temporal with helm chart configured with sql/Postgres persistent storage

### Describe the bug

Configured helm values file using example for [Postgres](https://github.com/temporalio/helm-charts/blob/main/charts/temporal/values/values.postgresql.yaml). After running helm deploy noticed that container for `temporal-schema-setup` is in a crash loop. Logs say `cannot connect to database <>. Database does not exist`

Looking at the helm chart, noticed that in [server-job.yaml](https://github.com/temporalio/helm-charts/blob/main/charts/temporal/templates/server-job.yaml) a `create-default-store` and `create-visibility-store` init containers should be created. 

Generated complete manifest with  `helm template -f my_values.yaml . ` and noticed that `initContainers` section is empty (contains `[]` only)

### Minimal Reproduction

Using example values file for [Postgres](https://github.com/temporalio/helm-charts/blob/main/charts/temporal/values/values.postgresql.yaml), generate deployment manifests with:
```
helm template . -f values/values.postgresql.yaml --debug > test.yaml
```
Inspect section `# Source: temporal/templates/server-job.yaml` which shows no `initContainers` were generated
```
# Source: temporal/templates/server-job.yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: release-name-temporal-schema-setup
  labels:
    app.kubernetes.io/name: temporal
    helm.sh/chart: temporal-0.43.0
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/version: 1.24.2
    app.kubernetes.io/component: database
    app.kubernetes.io/part-of: temporal
  annotations:
    "helm.sh/hook": pre-install
    "helm.sh/hook-weight": "0"
    "helm.sh/hook-delete-policy": hook-succeeded,hook-failed
spec:
  backoffLimit: 100
  template:
    metadata:
      name: release-name-temporal-schema-setup
      labels:
        app.kubernetes.io/name: temporal
        helm.sh/chart: temporal-0.43.0
        app.kubernetes.io/managed-by: Helm
        app.kubernetes.io/instance: release-name
        app.kubernetes.io/version: 1.24.2
        app.kubernetes.io/component: database
        app.kubernetes.io/part-of: temporal
    spec:
      
      restartPolicy: "OnFailure"
      initContainers:
          []
      containers:
[...]

PR #522, opened 2024-07-03
Setup for `ci install`.

robholland:
Adds a test so that `helm test` now checks that the cluster is healthy after the system is deployed.

Refactors server-job to remove the use of helm hooks which cause lot of users pain and don't work with --wait, or the mechanism that helm test uses. This should allow deployment to work when done via Terraform/Pulumi and other similar system which use a wait-style approach.

Improve the handling of elasticsearch by treating it just like the other drivers. Also uses a secret for Elasticsearch password rather than having it in plain text, inline with the other drivers.

Correctly handle the schema.createDatabase setting which previously had edge cases.

Note: This PR does not enable `ct install` to run in any PR or release workflows, but that will be done soon in another PR.

PR #524, opened 2024-07-03
[Issue 523] Fix logic for createDatabase.enabled variable

ogusak:
## What was changed
Fixed logic flow for `.Values.schema.createDatabase.enabled` to allow creating databases if they are not provisioned beforehand. 

## Why?
See bug described in https://github.com/temporalio/helm-charts/issues/523
## Checklist

1. Closes 523

2. How was this tested:
`helm template . -f values/values.postgresql.yaml --debug > test.yaml`
and inspecting content of the resulting manifest. Tested for 
- `.Values.schema.createDatabase.enabled=true`
- `.Values.schema.createDatabase.enabled=false`

3. Any docs updates needed?
N/A

### DIFF
--- a/charts/temporal/templates/server-job.yaml
+++ b/charts/temporal/templates/server-job.yaml
@@ -38,6 +38,7 @@ spec:
     spec:
       {{ include "temporal.serviceAccount" . }}
       restartPolicy: "OnFailure"
+      {{- if .Values.schema.createDatabase.enabled }}
       initContainers:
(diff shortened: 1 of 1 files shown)


robholland:
Should be fixed by #522, please test this and let me know.

robholland:
Please let me know if #522 does not fix this for you, I'll re-open.

ogusak:
Tested just now and can see `create-default-store` and `create-visibility-store` init containers are rendered. Thank you @robholland !

Issue #529, opened 2024-07-08
[Bug] check-elasticsearch-index init container on main deployment is not using ES secret for credentials


csabatuz-chess:
### What are you really trying to do?

We are trying to configure our in-house Temporal deployment with an external Elasticsearch, using authentication.

The credentials are put in the secret and key specified under:
```
elasticsearch:
  ...
  secretName: ...
  secretKey: ...
```

### Describe the bug

The `check-elasticsearch-index` init container on the main deployment fails due to credential errors.
https://github.com/temporalio/helm-charts/blob/main/charts/temporal/templates/server-deployment.yaml#L69-L75

#### Our own analysis

It appears that it expects the password under `elasticsearch.password` value in cleartext. 
While this is good for testing, in our actual deployment we do need to use it from the secret. 

It appears that jobs are using the secret as expected, via `_admintools-env.yaml`:
https://github.com/temporalio/helm-charts/blob/main/charts/temporal/templates/server-job.yaml#L44-L46

_It's possible that the Deployment template's current state is a leftover after a previous refactor_ 

Our naive fix attempt (untested for now): https://github.com/csabatuz-chess/temporal-helm-charts/compare/main..CST/elasticsearch-password-from-secret

### Minimal Reproduction

Stock charts with
```
elasticsearch:
  enabled: false
  external: true
  ... other configs ...
  username: elastic
  secretName: <your secret>
  secretKey: password
```

```
apiVersion: v1
data:
  password: <your password>
kind: Secret
```

### Environment/Versions

We are using Kubernetes, with 0.44.0 version of the chart.

robholland:
Fixed by: https://github.com/temporalio/helm-charts/pull/530

PR #530, opened 2024-07-08
Use admintools-env and secret for ES password consistent with e.g. jobs

csabatuz-chess:
## What was changed
The ES check init container on the main deployment of temporal server is made consistent with the Job templates' similar containers.

## Why?
Problem encountered with setting up Temporal deployment with external Elasticsearch using authentication.
Issue created:
https://github.com/temporalio/helm-charts/issues/529

## Why draft
We need to work a bit to test this chart version with our deployment


robholland:
We can test this also, the change looks roughly right to me, feel free to mark this PR as ready for review.

robholland (approved):

Issue #540, opened 2024-07-29
[Feature Request] Allow setting authorization via helm values


qs-synth:
### Is your feature request related to a problem? Please describe.

I would like to be able to set the [authorization section](https://github.com/temporalio/temporal/blob/459ad6902d59961a41a5657e9f55ff621805b1d9/docker/config_template.yaml#L256-L268) of `config_template.yaml` from helm values.

### Describe the solution you'd like

It would be simplest if I could pass in an `authorization` block in the same way [tls](https://github.com/temporalio/helm-charts/blob/40acb649b1276cb778e064bf6930aa6facdda344/charts/temporal/values.yaml#L92) is passed in the values file.

### Additional context

In my specific use-case, I would like to enable the default ClaimMapper and Authorizer without needing to rebuild the server image.

```
  config:
    authorization:
      jwtKeyProvider:
        keySourceURIs:
          - http://hostname:port/jwks.json
        refreshInterval: 1m
      permissionsClaimName: permissions
      authorizer: default
      claimMapper: default
```

PR #541, opened 2024-07-29
add authorizer field

qs-synth:
## What was changed
- added `authorization` block to templates to specify Authorizer and ClaimMapper settings

## Why?
This was done to allow overriding the noOp Authorizer and ClaimMapper when deploying via helm in the same way that the docker [config_template.yaml](https://github.com/temporalio/temporal/blob/main/docker/config_template.yaml#L256-L268) handles overriding these values.

## Checklist

1. Closes https://github.com/temporalio/helm-charts/issues/540

2. How was this tested:

TBD, working on this

3. Any docs updates needed?
I have updated the values file to include a commented example of the fields available. This matches what was implemented for `tls` field.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,27 @@
+name: Release Charts
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  release:
+    permissions:
+      contents: write
+    runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


qs-synth:
Closing in preference of a PR from an isolated branch: https://github.com/temporalio/helm-charts/pull/542

PR #542, opened 2024-07-29
Feat add authorization options

qs-synth:
## What was changed
- added `authorization` block to templates to specify Authorizer and ClaimMapper settings

## Why?
This was done to allow overriding the noOp Authorizer and ClaimMapper when deploying via helm in the same way that the docker [config_template.yaml](https://github.com/temporalio/temporal/blob/main/docker/config_template.yaml#L256-L268) handles overriding these values.

## Checklist

1. Closes https://github.com/temporalio/helm-charts/issues/540

2. How was this tested:

I have deployed a simple service based on the DefaultAuthorizer and [JWT issuer example](https://github.com/temporalio/samples-server/tree/main/extensibility/authorizer) within the same cluster as my helm deployment. Once deployed, I was able to issue a token with the steps from the JWT Issuer example and access temporal using [tctl](https://docs.temporal.io/tctl-v1) to list namespaces. 

The latest version of [Temporal CLI](https://docs.temporal.io/cli), does not appear to support passing an authorization token though.

3. Any docs updates needed?
I have updated the values file to include a commented example of the fields available. This matches what was implemented for `tls` field.


wrbbz:
> The latest version of [Temporal CLI](https://docs.temporal.io/cli), does not appear to support passing an authorization token though.

As I recall, you can pass it via `TEMPORAL_API_TOKEN` env var

robholland (approved):

Issue #532, opened 2024-07-16
[Bug] Schema Job not updating schema when using postgres12_pgx driver


Oro:
### What are you really trying to do?

Using `postgres12_pgx` driver instead of the `postgres` driver results in the pod running schema updates to stay in podinitializing and never executing the schema upgrade. 

### Describe the bug
Running (with an existing postgres schema in my case, but should be the same even on an empty db) the following will start the the schema job that stays indefinitely as running with the following full log output:
```
INFO	Schema setup complete	{"logging-call-at": "setuptask.go:149"}
DEBUG	Current database schema version 1.6 is greater than initial schema version 0.0. Skip version upgrade	{"logging-call-at": "setuptask.go:134"}
DEBUG	Setting up version tables	{"logging-call-at": "setuptask.go:73"}
INFO	Starting schema setup	{"config": {"SchemaFilePath":"","SchemaName":"","InitialVersion":"0.0","Overwrite":false,"DisableVersioning":false}, "logging-call-at": "setuptask.go:63"}
INFO	Schema setup complete	{"logging-call-at": "setuptask.go:149"}
DEBUG	Current database schema version 1.12 is greater than initial schema version 0.0. Skip version upgrade	{"logging-call-at": "setuptask.go:134"}
DEBUG	Setting up version tables	{"logging-call-at": "setuptask.go:73"}
INFO	Starting schema setup	{"config": {"SchemaFilePath":"","SchemaName":"","InitialVersion":"0.0","Overwrite":false,"DisableVersioning":false}, "logging-call-at": "setuptask.go:63"}
```

### Minimal Reproduction
```
  cassandra:
    enabled: false
  prometheus:
    enabled: false
  grafana:
    enabled: false
  elasticsearch:
    enabled: false
  postgresql:
    enabled: true
  schema:
    createDatabase:
      enabled: false
  server:
    config:
      persistence:
        default:
          driver: "sql"
          sql:
            driver: postgres12_pgx
            host: temporal-defaultstore-postgres
            port: 5432
            database: app
            user: app
            existingSecret: temporal-defaultstore-postgres-app
            maxConns: 20
            maxConnLifetime: "1h"
        visibility:
          driver: "sql"
          sql:
            driver: postgres12_pgx
            host: temporal-visibilitystore-postgres
            port: 5432
            database: app
            user: app
            existingSecret: temporal-visibilitystore-postgres-app
            maxConns: 20
            maxConnLifetime: "1h"
```
[...]

PR #546, opened 2024-08-05
Check for pgx driver as well as plain postgres.

robholland:
## What was changed
Check for pgx driver as well as plain postgres driver.

## Why?
The schema should be created/updated in either case.

## Checklist

1. Closes: https://github.com/temporalio/helm-charts/issues/532

2. How was this tested:

3. Any docs updates needed?


MasonEgger (approved):
LGTM

PR #550, opened 2024-08-27
Provide option to create default namespace

### LINKED ISSUE #420: [Feature Request] Add option to create default namespace

mananmangal:
## What was changed

This PR addresses the Feature Request #420 
Update the server setup Kubernetes job to create a default namespace.
Create configuration to enable this feature, by default its disabled.
Also, provides option to configure retention period, by default its 3 days.

## Why?

Creates a default namespace on the server to avoid manual effort.

## Checklist

1. Closes #420 

2. How was this tested:  Manual installation of the helm chart and running the updated job on a Kubernetes cluster to validate creation of new namespace.
`helm lint charts/temporal/.`
`helm install --dry-run temporal charts/temporal/.`
`helm install temporal charts/temporal/.`

4. Any docs updates needed? N/A

### DIFF
--- a/charts/temporal/values.yaml
+++ b/charts/temporal/values.yaml
@@ -178,11 +178,12 @@ server:
           maxConnLifetime: "1h"
           # connectAttributes:
           #   tx_isolation: 'READ-COMMITTED'
-    defaultNamespace:
-      # Enable this to create default namespace
+    defaultNamespaces:
+      # Enable this to create namespaces
       enabled: false
-      namespace: default
-      retention: 3d
+      namespace:
+        - name: default
+          retention: 3d
   frontend:
     service:
       # Evaluated as template
--- a/charts/temporal/values.yaml
+++ b/charts/temporal/values.yaml
@@ -178,9 +178,9 @@ server:
           maxConnLifetime: "1h"
           # connectAttributes:
           #   tx_isolation: 'READ-COMMITTED'
-    defaultNamespaces:
+    namespaces:
       # Enable this to create namespaces
-      enabled: false
+      create: false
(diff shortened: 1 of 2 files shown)


robholland (changes requested):
Please adjust to take a list of namespaces rather than just one. Probably as a list of maps, so they can set the namespace name and (optionally) retention per namespace.

trin4ik:
the right thing, I need it.

charts/temporal/values.yaml
```diff
@@ -178,6 +178,12 @@ server:
           maxConnLifetime: "1h"
           # connectAttributes:
           #   tx_isolation: 'READ-COMMITTED'
+    defaultNamespaces:
```
robholland, line 181:
```suggestion
    namespaces:
```

charts/temporal/values.yaml
```diff
@@ -178,6 +178,12 @@ server:
...
+      # Enable this to create namespaces
+      enabled: false
```
robholland, line 183:
```suggestion
      create: false
```

robholland (changes requested):

charts/temporal/values.yaml
mananmangal: Applied this change

charts/temporal/values.yaml
mananmangal: Applied this change

robholland (approved):

robholland:
I'm on the fence about whether this should happen only if we have `schema.create/update` turned on, but I can revisit that. Thanks for you contribution :)

PR #553, opened 2024-09-09
Restrict CIDR ip addresses for a LoadBalancer type service

asproul:
## What was changed

Restrict IP addresses for load balancers created for the frontend service by adding the `spec.loadBalancerSourceRanges` for the LoadBalancer type. 

## Why?

I am self-hosting temporal on a Google Kubernetes Engine (GKE) cluster running temporal workflows. I have exposed the frontend service as an internal load balancer to my Cloud Run service (which initiates the temporal workflows) but, by default, the firewall rule created by the GKE service sets a source as "0.0.0.0/0". I am required to further restrict the allowed source ip range to my Cloud Run subnet.

[GCP guide]https://cloud.google.com/kubernetes-engine/docs/concepts/service-load-balancer-parameters#fw_ip_address) about restricting ip addresses based on CIDRs in `spec.loadBalancerSourceRanges[]`. Additionally, [AWS documentation](https://repost.aws/knowledge-center/eks-cidr-ip-address-loadbalancer) states that `spec.loadBalancerSourceRanges[]` is used to restrict IP addresses. 

## Checklist

1. Closes 

2. How was this tested:
* Rendered the template locally.
* Deployed the changes to my development google kubernetes engine cluster. 
* Verified that the firewall rules concerning this frontend service changed the sourceIP from 0.0.0.0/0 to the ip range I specified in the `server.frontend.service.loadBalancerSourceRanges` from my helm values.yaml file.

3. Any docs updates needed?

### DIFF
--- a/charts/temporal/templates/server-service.yaml
+++ b/charts/temporal/templates/server-service.yaml
@@ -10,6 +10,9 @@ metadata:
   {{- end }}
 spec:
   type: {{ .Values.server.frontend.service.type }}
+  {{- if .Values.server.frontend.service.loadBalancerSourceRanges }}
+  loadBalancerSourceRanges: {{- include "common.tplvalues.render" ( dict "value" .Values.server.frontend.service.loadBalancerSourceRanges "context" $) | nindent 4 }}
+  {{- end }}
   ports:
     - port: {{ .Values.server.frontend.service.port }}
       targetPort: rpc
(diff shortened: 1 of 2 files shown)


robholland:
Could you add an example to the values file of what these entries would look like? Or if you're willing/able, add a test file for the service, like: https://github.com/temporalio/helm-charts/blob/main/charts/temporal/tests/server_deployment_test.yaml for the deployment.

asproul:
I messed up the merge so I closed out this PR in favor of a new one (https://github.com/temporalio/helm-charts/pull/614).

PR #555, opened 2024-09-10
Fix schema job not garbage collected

### LINKED ISSUE #554: [Bug] `helm upgrade` does not work if `schema.enabled` is true

### Describe the bug

### Minimal Reproduction
1. Install helm chart

### Environment/Versions

kosmoz:
## What was changed
This PR fixes #554 by allowing users to specify a `ttlSecondsAfterFinished`. 

## Why?
If the job is not garbage collected, helm upgrades are not possible.

## Checklist

1. Closes #554

2. How was this tested: 
   Run helm install, check if job is deleted after completion

4. Any docs updates needed? 
   Don't know, is there documentation for the helm chart?

### DIFF
--- a/charts/temporal/templates/server-job.yaml
+++ b/charts/temporal/templates/server-job.yaml
@@ -7,6 +7,9 @@ metadata:
     {{- include "temporal.resourceLabels" (list . "database" "") | nindent 4 }}
 spec:
   backoffLimit: {{ $.Values.schema.setup.backoffLimit }}
+  {{- if $.Values.schema.ttlSecondsAfterFinished }}
+  ttlSecondsAfterFinished: {{ $.Values.schema.ttlSecondsAfterFinished }}
+  {{- end }}
   template:
     metadata:
       name: {{ include "temporal.componentname" (list . "schema") }}
--- a/charts/temporal/values.yaml
+++ b/charts/temporal/values.yaml
@@ -379,6 +379,8 @@ schema:
   resources: {}
   containerSecurityContext: {}
   securityContext: {}
+  # To delete the Job resource after it is finished. This can help with subsequent upgrades
+  # ttlSecondsAfterFinished: 10
 elasticsearch:
   enabled: true
   replicas: 3


charts/temporal/templates/server-job.yaml
```diff
@@ -7,6 +7,9 @@ metadata:
     {{- include "temporal.resourceLabels" (list . "database" "") | nindent 4 }}
 spec:
   backoffLimit: {{ $.Values.schema.setup.backoffLimit }}
+  {{- if $.Values.schema.ttlSecondsAfterFinished }}
+  ttlSecondsAfterFinished: {{ $.Values.schema.ttlSecondsAfterFinished }}
```
joshbranham, line 11: We haven't hit this yet but now I see it being problematic for subsequent schema upgrades. Should we just make this a new default with a sane value like a day? Trying to imagine a scenario where you would want the job to exist after the fact, since the Pod (in our case, due to node rotations) is already gone.

kosmoz: I guess that's up for you to decide. We have set it very low (10 seconds) because, especially if some sort of automation is used, updates can happen quite frequently, e.g. after configuration changes.

A more idiomatic approach would be using [chart hooks](https://helm.sh/docs/topics/charts_hooks/), but that doesn't work in this case, because all the deployments wait for the schema to be present but the post-install hook is only executed after deployed resources are ready, which leads to a deadlock situation.

FurqanHabibi: There used to be a `pre-install` chart hook on this job which was removed recently in this PR
https://github.com/temporalio/helm-charts/pull/522

I agree with @kosmoz that the chart hook is a better approach, and the `pre-install` hook works perfectly for this.
I would argue for re-introduction of this hook, maybe behind a flag so that people can choose whether they want the job as hook or not.

kosmoz: As long as it works, that should be fine by me :-)

robholland: As detailed in the PR that removes the hooks, they can break anything which uses helm in wait mode, including terraform, pulumi, Argo and somewhat ironically, `helm test`. I will look at this again, but I doubt we'll be able to put hooks back in.

robholland: Please test #564 as a solution to this issue.

robholland:
Fixed by #564.

charts/temporal/templates/server-job.yaml
martin31821: @robholland I've just run into the same issue with argo, I'd like to chime in with a potential idea:

- Allow the user to set ttlSeconds to null, and if null, don't set it on the hook itself
- Allow the user to set custom annotations on the job, allowing users to use argo/CD native hook functionality:

```yaml
argocd.argoproj.io/hook: Sync
argocd.argoproj.io/hook-delete-policy: BeforeHookCreation
argocd.argoproj.io/sync-wave: 'x'
```
We're using temporal as part of peerdb and this would solve both cases.

PR #559, opened 2024-09-19
Fix schema job cleanup by adding optional hook annotation

### LINKED ISSUE #554: [Bug] `helm upgrade` does not work if `schema.enabled` is true

FurqanHabibi:
## What was changed

This PR adds optional chart hook annotations to schema job. This `pre-install` hook will make sure the job gets run before any other deployment, and will automatically cleans up the job after it's finished with a `hook-delete-policy`.

## Why?

If the schema job is not garbage collected helm upgrades are not possible.
Also if the job doesn't get run first, other deployments can fail (at least initially) until the job is finished.

## Checklist

1. Closes #554 

2. How was this tested:
  Run helm install, check the job runs first before any other deployment, and check the job gets deleted after completion.

3. Any docs updates needed?
  I don't think there is any doc for this.

### DIFF
--- a/charts/temporal/templates/server-job.yaml
+++ b/charts/temporal/templates/server-job.yaml
@@ -5,6 +5,12 @@ metadata:
   name: {{ include "temporal.componentname" (list . "schema") }}
   labels:
     {{- include "temporal.resourceLabels" (list . "database" "") | nindent 4 }}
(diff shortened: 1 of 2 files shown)


robholland:
Fixed by #564

Issue #464, opened 2024-02-08
[Feature Request] [Archival] [s3 provider] use IAM role + serviceAccount


luismacosta:
### Is your feature request related to a problem? Please describe.
[Archival] [s3 provider] use IAM role + serviceAccount

### Describe the solution you'd like
- use IAM role + IAM policy
- create serviceAccount with annotation _eks.amazonaws.com/role-arn_

### Additional context
I've tried to configure this, but got an error regarding s3 access
From what I could find in the community, at the moment it is only possible to use IAM user (env vars AWS_ACCESS_KEY_ID + AWS_SECRET_ACCESS_KEY)

disconnex:
I would love to see this feature as well, since it makes sense in an EKS environment to use service accounts and IAM roles over creating users and access keys.

PR #562, opened 2024-09-24
Fix service account creation when no Values.serviceAccount.name is defined

luong-komorebi:
## What was changed

Instead of checking for `serviceAccount.name` inside a helper template, I rely on `serviceAccount.create`and ensure a service account is always used instead of the service account `default`

## Why?

Given a yaml values:

```yaml
serviceAccount:
  create: true
```

Without this change, a service account is created with the default name (which is taken from release full name as documented in[ values yaml ](https://github.com/luong-komorebi/temporal-helm-charts/blob/06e89b7f986ba4afcab2126ef1d61eceffb39d5c/charts/temporal/values.yaml#L11)) but this new service account will never be used because the helper template will never refer to it 

Instead all the pods created by temporal will always use `serviceAccountName: default` and default will come with no configuration, such as annotations that people use on popular k8s platform like eks 

```yaml
              extraAnnotations:
                eks.amazonaws.com/role-arn: <some arn here>
```

Just because of a missing serviceaccount name, any other configuration for service account will never be applied 

### Extra context
I figured out this issue when trying to debug why archivals did not work and using temporal operator to enable archivals is giving back 403 in the server log 
After enabling debugging log for archival s3store, I found 

```
temporal-frontend -----------------------------------------------------
temporal-frontend 2024/09/24 08:18:47 DEBUG: Response s3/HeadBucket Details:
temporal-frontend ---[ RESPONSE ]--------------------------------------
temporal-frontend HTTP/1.1 403 Forbidden
temporal-frontend Connection: close
temporal-frontend Content-Type: application/xml
temporal-frontend Date: Tue, 24 Sep 2024 08:18:46 GMT
temporal-frontend Server: AmazonS3
temporal-frontend X-Amz-Bucket-Region: us-west-2
temporal-frontend X-Amz-Id-2: <stripped>
temporal-frontend X-Amz-Request-Id: GEQS26YQT7Q0WSCM
temporal-frontend
temporal-frontend
temporal-frontend -----------------------------------------------------
```

which hints at the permission issue with the role. 

## Checklist

1. Closes  #464

2. How was this tested: Manually on my platform. New pods will be created with 

```yaml
      serviceAccount: temporal
      serviceAccountName: temporal
```
 

4. Any docs updates needed? No


charts/temporal/templates/_helpers.tpl
```diff
@@ -42,7 +42,7 @@ Create the name of the service account
 Define the service account as needed
 */}}
 {{- define "temporal.serviceAccount" -}}
-{{- if .Values.serviceAccount.name -}}
+{{- if .Values.serviceAccount.create -}}
```
robholland, line 45: Makes sense, but we need to still use the service account if `create` is false but `name` is set.

robholland (changes requested):

charts/temporal/templates/_helpers.tpl
robholland: This prevents users from using a pre-made service account. Helm will fail trying to create the serviceAccount that already exists.

robholland: Sorry, I don't understand your fix. Now we might create a serviceAccount called "default". I don't understand the reasoning there. Please see if https://github.com/temporalio/helm-charts/pull/574 fixes the behaviour you needed.

charts/temporal/values.yaml
```diff
@@ -8,7 +8,7 @@ imagePullSecrets: {}
 serviceAccount:
   # Whether to create service account or not
   create: false
-  # Name of the service account, default: temporal.fullname
+  # Name of the service account, default if `create: true` and `name` not defined: temporal.fullname
```
robholland, line 11:
```suggestion
  # Name of the service account.
  # Uses `name` if set.
  # If `create: true` defaults to temporal.fullname, otherwise uses the namespace's default serviceAccount.
```

robholland (approved):
Thank you for your explanation and patience :) This all makes sense to me now, I've tried to tweak the comment a little, but otherwise looks good!

robholland (approved):

PR #564, opened 2024-09-25
Create a job per chart version so that schema upgrades can run.

robholland:
## What was changed

Added the chart version into the name of the schema job. Set a TTL so that the job and its resources are cleaned up after 24 hours.

## Why?

This avoids job creation failing when doing upgrades.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


robholland:
Alternative to https://github.com/temporalio/helm-charts/pull/559 and https://github.com/temporalio/helm-charts/pull/555.

prathyushpv (approved):

PR #566, opened 2024-09-26
Config to specify tags to be excluded in prometheus metrics

prathyushpv:
## Why?
Config to exclude metrics tags from prometheus.

## Checklist

1. Closes

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/charts/temporal/templates/server-configmap.yaml
+++ b/charts/temporal/templates/server-configmap.yaml
@@ -88,10 +88,10 @@ data:
           {{- with $server.metrics.tags }}
             {{- toYaml . | nindent 10 }}
           {{- end }}
+        {{- with $server.metrics.excludeTags }}
         excludeTags:
-          {{- with $server.metrics.excludeTags }}
-            {{- toYaml . | nindent 10 }}
-          {{- end }}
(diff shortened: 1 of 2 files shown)


charts/temporal/templates/server-configmap.yaml
```diff
@@ -88,6 +88,10 @@ data:
           {{- with $server.metrics.tags }}
             {{- toYaml . | nindent 10 }}
           {{- end }}
+        excludeTags:
+          {{- with $server.metrics.excludeTags }}
+            {{- toYaml . | nindent 10 }}
+          {{- end }}
```
robholland, line 94:
```suggestion
        {{- with $server.metrics.excludeTags }}
        excludeTags:
          {{- toYaml . | nindent 10 }}
        {{- end }}
```

robholland: Avoid adding an empty `excludeTags` block.

robholland (changes requested):

prathyushpv:
Thanks Rob! I have applied the suggested change.

robholland (approved):

Issue #560, opened 2024-09-22
[Feature Request] Ability to specify internal frontend in Helm chart


Meeki1:
Since version 1.20 it became possible to use internal frontend. It would be nice if it was possible to use it in helm chart too

mxk:
It looks like support for internal-frontend has been added. However, the web UI still uses [frontend](https://github.com/temporalio/helm-charts/blob/c0eaf12c681d78da5215d2f85c17d3c73ff702dd/charts/temporal/templates/web-deployment.yaml#L34-L35). Should that be updated to use internal frontend when that is enabled?

robholland:
No, I don't believe it's intended for the web frontend to use the internal frontend.

Closing this now #602 is merged.

PR #571, opened 2024-09-30
Including an internal-frontend service with auth enabled

dleblanc:
NOTE: I have not tested this with mTLS.

## What was changed
This adds an internal-frontend portion to the config map when a non-default authorizer is enabled, and removes the publicClient section if it's present.

It starts the internal-frontend deployment using the same parameters as the normal frontend, but slightly different ports (7236, 6936).

This pretty much follows the instructions in the release notes for 1.20: https://github.com/temporalio/temporal/releases/tag/v1.20.0

## Why?
When one enables authorization using the recent jwt authorization support, the worker pod fails repeatedly with an authorization failure because it is not providing the JWT token.  By following the above instructions, we bring up a special "internal frontend" that it will use for internal communications instead, bypassing this requirement and allowing the worker to start up properly.

1. Closes issue #560 

2. How was this tested:

Added the requisite configuration, and ensure that the Temporal worker starts successfully.

3. Any docs updates needed?

Added a section to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,6 +316,36 @@ helm install \
   --wait
 ```
 
+### Configuring authorization
+
+See https://docs.temporal.io/self-hosted-guide/security#claim-mapper
+
+This chart enables you to configure Temporal's authorization support. The values within this Helm chart
+server.config.authorization section are included in the resulting Temporal configuration under the 
+global.authorization section.
+
+For instance, you could enable the JWT based default authorizer and claim mapper functionality by including the following:
+
+This configures Temporal to load the JSON Web Key Set from the provided keySourceURIs, and refresh it on an internal. 
+That is used to validate the JWTs that you provide to Temporal via the "authorization" meta data header. 
+
+```yaml
+server:
+  config:    
+    authorization:
+     jwtKeyProvider:
+       keySourceURIs:
+         - http://localhost:/jwks.json
+       refreshInterval: 1m
+     permissionsClaimName: permissions
+     authorizer: default
+     claimMapper: default
+```
+
+Note that if a non-default value is provided for server.config.authorization.authorizer, then we will automatically
+include the "internal-frontend" service that the other Temporal services use for internal communication that bypasses the 
+authorization process. 
+
 ## Play With It
 
 ### Exploring Your Cluster
--- a/charts/temporal/templates/server-configmap.yaml
+++ b/charts/temporal/templates/server-configmap.yaml
@@ -119,6 +119,15 @@ data:
           membershipPort: {{ $server.frontend.service.membershipPort }}
           bindOnIP: "0.0.0.0"
 
+      {{- if and (hasKey .Values.server.config "authorization") (hasKey .Values.server.config.authorization "authorizer") }}
+      internal-frontend:
+        rpc:
+          grpcPort: {{ index $server "internal-frontend" "service" "port" }}
+          httpPort: {{ index $server "internal-frontend" "service" "httpPort" }}
+          membershipPort: {{ index $server "internal-frontend" "service" "membershipPort" }}
+          bindOnIP: "0.0.0.0"
+      {{- end }}
+
       history:
         rpc:
           grpcPort: {{ $server.history.service.port }}
@@ -173,8 +182,10 @@ data:
       {{- toYaml . | nindent 6 }}
     {{- end }}
 
+    {{- if not (and (hasKey .Values.server.config "authorization") (hasKey .Values.server.config.authorization "authorizer")) }}
     publicClient:
       hostPort: "{{ include "temporal.componentname" (list $ "frontend") }}:{{ $server.frontend.service.port }}"
+    {{- end }}
 
     dynamicConfigClient:
       filepath: "/etc/temporal/dynamic_config/dynamic_config.yaml"
--- a/charts/temporal/templates/server-deployment.yaml
+++ b/charts/temporal/templates/server-deployment.yaml
@@ -1,5 +1,12 @@
 {{- if $.Values.server.enabled }}
-{{- range $service := (list "frontend" "history" "matching" "worker") }}
+
+{{- $serviceList := (list "frontend" "history" "matching" "worker") }}
+
+{{- if and (hasKey .Values.server.config "authorization") (hasKey .Values.server.config.authorization "authorizer") }}
+  {{- $serviceList = (list "frontend" "internal-frontend" "history" "matching" "worker") }}
+{{- end }}
+
+{{- range $service := $serviceList }}
 {{ $serviceValues := index $.Values.server $service }}
 apiVersion: apps/v1
 kind: Deployment
--- a/charts/temporal/values.yaml
+++ b/charts/temporal/values.yaml
@@ -218,6 +218,49 @@ server:
       #  - secretName: chart-example-tls
       #    hosts:
       #      - chart-example.local
+
+    metrics:
+      annotations:
+        enabled: true
+      serviceMonitor: {}
+      # enabled: false
+      prometheus: {}
+      # timerType: histogram
+    podAnnotations: {}
+    podLabels: {}
+    resources: {}
+    nodeSelector: {}
+    tolerations: []
+    affinity: {}
+    additionalEnv: []
+    containerSecurityContext: {}
+    topologySpreadConstraints: []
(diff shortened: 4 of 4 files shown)


adamko147:
I would have a suggestion, why to decide based on authorization settings? could we simplify with settings

```yaml
server:
    frontend:
        enabled: false
        # other config for frontend component
    internalFrontend:
        enabled: true
        # other config for internal-frontend component
```
and do the implementation based on `{{ .Values.server.frontend.enabled }}` and `{{ .Values.server.internalFrontend.enabled }}`
in our scenario, we have our custom frontend build with authorizer independent of the `server.config.authorization` value, so your solution would not work for us without tweaking the helm char. 

Thank you

dleblanc:
> I would have a suggestion, why to decide based on authorization settings? could we simplify it with settings
> 
> ```yaml
> server:
>     frontend:
>         enabled: false
>         # other config for frontend component
>     internalFrontend:
>         enabled: true
>         # other config for internal-frontend component
> ```
> 

Yeah, I like that approach better. I would still need to know when to remove the publicClient stuff, I can probably add another config value for that.

I'll see about putting those changes in, ideally in the next few days.

adamko147:
> 
> Yeah, I like that approach better. I would still need to know when to remove the publicClient stuff, I can probably add another config value for that.

I think the publicClient would be required when `frontend.enabled == true`, assuming if you're deploying built-in frontend (not your own), you want that to be accessible by clients.

TheGeniesis:
Hi,
You read in my mind - I have similar scenario - I want to use service principals and Azure Entra ID to handle users and apps auth.
Since, you already created MR I want to highlight some changes which I did to make everything work:

> The source version was 0.44.0.
> I was doing my changes to check if it works in internal project, before I thought it might be good to raise MR to original repo.

## Change 1
I had to add `if` block in [server-configmap.yaml](https://github.com/temporalio/helm-charts/blob/main/charts/temporal/templates/server-configmap.yaml#L109). Only frontend is used externally, so only this service should have oAuth enabled (otherwise everything fails).

```yaml
      {{- if eq $service "frontend" }}
      {{- with $server.config.authorization }}
      authorization:
        {{- toYaml . | nindent 10 }}
      {{- end }}
      {{- end }}
```

## Change 2

Repo uses camelCase, so I used `internalFrontend` in values.yaml and added in [_helpers.tpl](https://github.com/temporalio/helm-charts/blob/main/charts/temporal/templates/_helpers.tpl):
```yaml
{{- define "serviceName" -}}
    {{- $service := index . 0 -}}
    {{- if eq $service "internalFrontend" }}
        {{- print "internal-frontend" }}
    {{- else }}
        {{- print $service }}
    {{- end }}
{{- end -}}
```
## Change 3

Usage for service-configmap.yaml:

```yaml
{{- range $originalService := (list "frontend" "internalFrontend" "history" "matching" "worker") }}
{{ $serviceValues := index $.Values.server $originalService }}
{{ $service := include "serviceName" (list $originalService) }}
apiVersion: v1
kind: ConfigMap
metadata:
  name: "{{ include "temporal.fullname" $ }}-config-{{ $service }}"
```

- `serviceName` might not be the best name here
- `internalFrontend`  should be added regarding to the condition in commend above
-   I added `internalFrontend` to all places where `frontend` was used (made a loop when needed)

## Change 4

Later in service-configmap.yaml I added missing section (condition required):
```yaml
      {{- if ne $service "frontend"}}
      internal-frontend:
        rpc:
            grpcPort: {{ $server.internalFrontend.service.port }}
            httpPort: {{ $server.internalFrontend.service.httpPort }}
            membershipPort: {{ $server.internalFrontend.service.membershipPort }}
            bindOnIP: "0.0.0.0"
      {{- end }}
```

## Change 5

Values.yaml

```yaml
  internalFrontend:
    service:
      annotations: {}
      type: ClusterIP
      port: 7236
      membershipPort: 6936
      httpPort: 7246
    ingress:
      enabled: false
      annotations: {}
      hosts:
        - "/"
      tls: []
    metrics:
      annotations:
        enabled: true
      serviceMonitor: {}
      prometheus: {}
    podAnnotations: {}
    podLabels: {}
    resources: {}
    nodeSelector: {}
    tolerations: []
    affinity: {}
    additionalEnv: []
    containerSecurityContext: {}
    topologySpreadConstraints: []
    podDisruptionBudget: {}
```

## Change 6 

To run admintool commands inside pod I have to add Authorization header:

```bash
temporal operator namespace create --namespace <namespace> --grpc-meta=Authorization='Bearer <token_from_ui>'
```

qs-synth:
I had a similar issue after enabling authorizer and have enabled internal-frontend by adding an extra service endpoint rather than an entirely new pod. I did it that way because I found that the frontend was throwing errors on startup when I had specified `publicClient` alongside `services.internal-frontend` and the configurations are mutually exclusive. The way I worked around it only takes a few changes.

For the deployment, pass env SERVICES="frontend:internal-frontend" to frontend container env
```
          {{- if and (eq $service "frontend") ($.Values.server.frontend.internal.enabled) }}
            - name: SERVICES
              value: "{{ $service }}:internal-frontend"
          {{- else }}
            - name: SERVICES
              value: {{ $service }}
          {{- end }}
```

Plus the extra port
```
          ports:
          ...
          {{- if and (eq $service "frontend") ($.Values.server.frontend.internal.enabled) }}
            - name: rpc-internal
              containerPort: {{ $serviceValues.internal.service.port }}
              protocol: TCP
          {{- end }}
```

For the configmap, internal-frontend looks the same as you have
```
      {{- if $server.frontend.internal.enabled }}
      internal-frontend:
        rpc:
          grpcPort: {{ $server.frontend.internal.service.port }}
          membershipPort: {{ $server.frontend.internal.service.membershipPort }}
          bindOnIP: "0.0.0.0"
      {{- end }}
```

But, `publicClient` has to be removed when internal is used. Prevents frontend from throwing startup errors.
```
    {{- if $server.frontend.internal.enabled}}
    {{- else }}
    publicClient:
      hostPort: "{{ include "temporal.componentname" (list $ "frontend") }}:{{ $server.frontend.service.port }}"
    {{- end }}
```

My values file addition looks like this. Same defaults used as the compose setup this is all based on.
```
  frontend:
    ...
    # Enables internal-frontend so that the builtin worker can access the frontend while
    # bypassing authorizer and claim mapper.
    # Equivalent to env.USE_INTERNAL_FRONTEND in docker compose config
    internal:
      enabled: false
      service:
        # Evaluated as template
        annotations: {}
        type: ClusterIP
        port: 7236
        membershipPort: 6936
```

adamko147:
> There appears to be another PR to accomplish this here as well: #602

I like the #602, it's kind of the implementation I had in mind. @dleblanc would #602 cover your use case as well? if yes, I would prefer to give it a "push" :) 

Thanks

dleblanc:
> > There appears to be another PR to accomplish this here as well: #602
> 
> I like the #602, it's kind of the implementation I had in mind. @dleblanc would #602 cover your use case as well? if yes, I would prefer to give it a "push" :)
> 
> Thanks

Yeah, I just ran it locally and it seems to meet my needs nicely. I'll close this PR in preference of #602.

alexku7:
Looks like the https://github.com/temporalio/helm-charts/pull/602 didn't resolve the issue completely

If I enable the JWT auth in the helm chart,  it's applied on the frontend and on the internal-fronted as well :(

TheGeniesis:
> Looks like the #602 didn't resolve the issue completely
> 
> If I enable the JWT auth in the helm chart, it's applied on the frontend and on the internal-fronted as well :(

Probably this scenario: https://github.com/temporalio/helm-charts/pull/602#discussion_r1820542381

There should be a condition in configmap to skip auth section for internalFrontend

alexku7:
> > Looks like the #602 didn't resolve the issue completely
> 
> > 
> 
> > If I enable the JWT auth in the helm chart, it's applied on the frontend and on the internal-fronted as well :(
> 
> 
> 
> Probably this scenario: https://github.com/temporalio/helm-charts/pull/602#discussion_r1820542381
> 
> 
> 
> There should be a condition in configmap to skip auth section for internalFrontend

Exactly but not only the auth section.
We also need an ability to set a different set if the tls options
The idea is to enable the tls client auth for the internal frontend and jwt auth for external frontend.
This is a very common scenario when you need a different set of options for external and internal frontend server

PR #573, opened 2024-10-01
Fix server-pdb

### LINKED ISSUE #572: [Bug] server-pdb is broken

Deploy temporal with pdb.

### Describe the bug

server-pdb is broken and can't be deployed because it uses `.` inside range block
https://github.com/temporalio/helm-charts/blob/main/charts/temporal/templates/server-pdb.yaml#L10

Instead it should use `$`.

### Minimal Reproduction

Try to deploy temporal with pdb
```
helm install \
    --repo https://go.temporal.io/helm-charts \
    --set server.frontend.replicaCount=2 \
    --set server.frontend.podDisruptionBudget.minAvailable=1 \
    temporaltest temporal \
    --timeout 15m
```

### Environment/Versions

- OS and processor: Linux
- Temporal Version: 1.25.0
- Kubernetes 1.30

FurqanHabibi:
## What was changed

Changed `.` to `$` in server-pdb range block since `.` in range [doesn't refer to top level scope](https://github.com/helm/helm/issues/3684).

## Why?

Currently server-pdb is broken and not deployable.
image

## Checklist

1. Closes #572

2. How was this tested:
    Tested deploying with pdb enabled
    ```
    helm install \
        --repo https://go.temporal.io/helm-charts \
        --set server.frontend.replicaCount=2 \
        --set server.frontend.podDisruptionBudget.minAvailable=1 \
        temporaltest temporal \
        --timeout 15m
    ```

3. Any docs updates needed?
    No

### DIFF
--- a/charts/temporal/templates/server-pdb.yaml
+++ b/charts/temporal/templates/server-pdb.yaml
@@ -7,7 +7,7 @@ kind: PodDisruptionBudget
 metadata:
   name: {{ include "temporal.componentname" (list $ $service) }}-pdb
   labels:
-    {{- include "temporal.resourceLabels" (list . $service "") | nindent 4 }}
+    {{- include "temporal.resourceLabels" (list $ $service "") | nindent 4 }}
 spec:
   {{ toYaml $serviceValues.podDisruptionBudget }}
   selector:


robholland:
This was fixed in https://github.com/temporalio/helm-charts/pull/568

PR #537, opened 2024-07-24
Add support for deployment labels and annotations

jonasrmichel:
## What was changed

This PR adds support for configuring custom labels and annotations in the main metadata of all `Deployment` workloads:
- `server`
- `server.frontend`
- `server.history`
- `server.matching`
- `server.worker`
- `admintools`
- `web`

## Why?

These changes were added specifically to allow for custom `server` `Deployment` annotations so that a tool like [stakater/Reloader](https://github.com/stakater/Reloader) may be used to auto-restart the `server` when persistence auth keys are rotated.

## Checklist

1. Closes 

2. How was this tested:

Local `helm template` rendering.

3. Any docs updates needed?

PR #582, opened 2024-10-03
Add support for deployment annotations

mark-hofmeijer:
## What was changed
This PR adds support to define deployment annotations for the `server`, `admintools` and `web` components. The setup is similar to the recent implemented resource labels.

NOTE: this PR partly replaces #537, however this PR needs revision because of the templatization in PR #539.

## Why?
Currently there is no option to define deployment annotations.

## Checklist
1. How was this tested:
`helm install -f values.yaml --set server.deploymentAnnotations.deployAnnotation1="serverDeploymentAnnotation" --set admintools.deploymentAnnotations.deployAnnotation1="adminDeploymentAnnotation" --set web.deploymentAnnotations.deployAnnotation1="webDeploymentAnnotation" debug . --dry-run --debug`

- In the current situation, this adds no deployment annotations on any of the components.
- In the new situation, a deployment annotation `deployAnnotation1` is added to all server services and admintools and web components.

`helm install -f values.yaml --set server.deploymentAnnotations.serverAnnotation1="serverDeploymentAnnotation" --set server.frontend.deploymentAnnotations.resourceAnnotation1="resourceDeploymentAnnotation" debug . --dry-run --debug`

- In the current situation, this adds no deployment annotations on the server `frontend` service.
- In the new situation, a deployment annotation `resourceAnnotation1` is added to the server `frontend` service. The deployment annotations on server level are ignored correctly.
- In the new situation, a deployment annotation `serverAnnotation1` is added to the other server services (`history`, `matching`, `worker`). 

2. Any docs updates needed?
No.


charts/temporal/templates/web-deployment.yaml
```diff
@@ -3,6 +3,10 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: {{ include "temporal.componentname" (list $ "web") }}
+  {{- if or $.Values.additionalAnnotations $.Values.admintools.deploymentAnnotations }}
```
robholland, line 6:
```suggestion
  {{- if or $.Values.additionalAnnotations $.Values.web.deploymentAnnotations }}
```

charts/temporal/templates/_helpers.tpl
```diff
@@ -71,6 +71,41 @@ Define the AppVersion
...
+{{- $resourceAnnotations := merge $scopeAnnotations $componentAnnotations -}}
+{{- range $annotation_name, $annotation_value := $resourceAnnotations }}
+{{ $annotation_name }}: {{ $annotation_value }}
+{{- end -}}
+{{- end -}}
+{{ include "temporal.additionalResourceAnnotations" $global }}
+{{- end -}}
+
+{{/*
+Additonal user specified annotations for all resources
```
robholland, line 101: Couldn't this just be inlined above? It's never called on it's own.

mark-hofmeijer:
Refactored the helper method to the new approach introduced in #584. Also added a test.

Noted that a new line is added in front of the annotations (both on deployment and pod level), but did not manage to remove it although the setup looks the same as for labels. Maybe you can take a quick look @robholland ?

charts/temporal/templates/admintools-deployment.yaml
```diff
@@ -3,6 +3,10 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: {{ include "temporal.componentname" (list $ "admintools") }}
+  {{- if or $.Values.additionalAnnotations $.Values.admintools.deploymentAnnotations }}
```
robholland, line 6: Let's remove the gating, we can just allow `annotations:` to be empty, it's valid syntax.

charts/temporal/templates/web-deployment.yaml
robholland: On seconds thoughts lets remove the gating.

robholland (changes requested):

charts/temporal/templates/_helpers.tpl
robholland: Just noticed the label version is split out. That should be inlined also, but don't worry about that for this PR.

charts/temporal/tests/deployment_test.go
```diff
@@ -104,3 +104,34 @@ func TestTemplateServerDeploymentMerging(t *testing.T) {
...
+	}
+
+	output := helm.RenderTemplate(t, options, helmChartPath, releaseName, []string{"templates/server-deployment.yaml"})
+
+	helm.UnmarshalK8SYaml(t, output, &deployment)
+
+	require.Equal(t, "three", deployment.ObjectMeta.Annotations["one"])
+	require.Equal(t, "two", deployment.ObjectMeta.Annotations["two"])
+	require.Equal(t, "four", deployment.ObjectMeta.Annotations["four"])
+}
```
robholland, line 137: 👍

charts/temporal/templates/_helpers.tpl
mark-hofmeijer: It was indeed intended to align with the approach on the labels. But agree that inline is better, because it's not called on it's own.

charts/temporal/tests/deployment_test.go
mark-hofmeijer: Did add a value test on the additional annotations as well.

charts/temporal/templates/web-deployment.yaml
mark-hofmeijer: Removed the gating.

charts/temporal/templates/admintools-deployment.yaml
mark-hofmeijer: Removed the gating.

robholland:
Please merge master to get the CI fix.

PR #584, opened 2024-10-07
Ensure we allow component level labels as well as scoped.

robholland:
Favour scoped labels.

## What was changed
Added some tests for server deployment template. Add support for component level resource labels. Favour scoped level resource labels.

## Why?
Component level labels were supported but accidentally removed in a PR. Put them back, and add some tests to ensure we don't regress again.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


axfelix (approved):
LGTM and a `helm test` is passing. However, I'm honestly not sure how I'm supposed to bootstrap a regular `go test` in this repo -- it's failing on missing repo sources -- and I'm not very familiar with terratest, but their docs imply I should be able to do that.

PR #585, opened 2024-10-08
feat: Use HEAD request to check if es_visibility index exists

### LINKED ISSUE #583: [Bug] Failed to validate ES index due to PUT request usage

I'm trying to upgrade Temporal helm chart in my environment `0.44.0 -> 0.46.2` (but I believe 0.47.0 is the same)
But the schema updated failing on ES index creation/check
image

### Describe the bug

The container `setup-visibility-store` fails after successful schema creation because of `PUT` method usage.
https://github.com/temporalio/helm-charts/blob/9ab3e8252b73cda773901336d3cae661a96cd6db/charts/temporal/templates/server-job.yaml#L80

Manual recreation:
```
 ~  curl -X PUT --fail --user "$ES_USER:$ES_PWD" $ES_SCHEME://$ES_HOST:$ES_PORT/_template/temporal_visibility_v1_template -H "Content-Type: application/json" --data-binary "@index_template_$ES_VERSION.json"
{"acknowledged":true}%

 ~  curl -X PUT --fail --user "$ES_USER:$ES_PWD" $ES_SCHEME://$ES_HOST:$ES_PORT/$ES_VISIBILITY_INDEX                                                                                                                                          
curl: (22) The requested URL returned error: 400
```

Here I want to clarify that I'm using AWS OpenSearch which is not supported by Temporal officially I guess?
```
  "version" : {
    "distribution" : "opensearch",
    "number" : "2.13.0",
```

Anyway I don't understand why we're using `-X PUT` here as the schema creation will be always `-X PUT`ed in the previous command (line 79):
https://github.com/temporalio/helm-charts/blob/9ab3e8252b73cda773901336d3cae661a96cd6db/charts/temporal/templates/server-job.yaml#L79

---

**Preferred Solution**

I see that both Elasticsearch and Opensearch support the following requests types to validate index existence:
1. `GET /index` - [ES docs](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/indices-get-index.html) | [OS docs](https://opensearch.org/docs/2.13/api-reference/index-apis/get-index/)
    ```
    ~ curl -X GET --fail --user "$ES_USER:$ES_PWD" $ES_SCHEME://$ES_HOST:$ES_PORT/$ES_VISIBILITY_INDEX                                                                                                                                          
    
    {"temporal_visibility_v1_dev":{"aliases":{},"mappings":{"dynamic":"false","properties": .......... [OUTPUT TRIMMED]
    ```
2. `HEAD /index` - [ES docs](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/indices-exists.html) | [OS docs](https://opensearch.org/docs/2.13/api-reference/index-apis/exists/)

    ```
     ~  curl --head --fail --user "$ES_USER:$ES_PWD" $ES_SCHEME://$ES_HOST:$ES_PORT/$ES_VISIBILITY_INDEX                                                                                                                                      
     
    HTTP/2 200
    date: Mon, 07 Oct 2024 11:12:50 GMT
    content-type: application/json; charset=UTF-8
    content-length: 1499
    access-control-allow-origin: *
    ```

Maybe it will make sense to rely on this types instead of PUT?

### Environment/Versions

- Temporal Version: temporaHelm-chart version `0.46.2`
- Kubernetes
- Visibility storage: `AWS OpenSearch 2.13 (ES v7 compatible)`

VLZZZ:
## What was changed
`curl` is now using `HEAD` request to ensure index existence and will result in 200/404 depending on the the result.
I decided to propose `HEAD` instead of `GET` as I believe we don't need all of the output `GET /index` will produce. See Context below.

**Context**
Both Elasticsearch and Opensearch support the following requests types to validate index existence:
1. `GET /index` - [ES docs](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/indices-get-index.html) | [OS docs](https://opensearch.org/docs/2.13/api-reference/index-apis/get-index/)
    ```
    ~ curl -X GET --fail --user "$ES_USER:$ES_PWD" $ES_SCHEME://$ES_HOST:$ES_PORT/$ES_VISIBILITY_INDEX                                                                                                                                          
    
    {"temporal_visibility_v1_dev":{"aliases":{},"mappings":{"dynamic":"false","properties": .......... [OUTPUT TRIMMED]
    ```
2. `HEAD /index` - [ES docs](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/indices-exists.html) | [OS docs](https://opensearch.org/docs/2.13/api-reference/index-apis/exists/)

    ```
     ~  curl --head --fail --user "$ES_USER:$ES_PWD" $ES_SCHEME://$ES_HOST:$ES_PORT/$ES_VISIBILITY_INDEX                                                                                                                                      
     
    HTTP/2 200
    date: Mon, 07 Oct 2024 11:12:50 GMT
    content-type: application/json; charset=UTF-8
    content-length: 1499
    access-control-allow-origin: *
    ```

## Why?
Please see  #583

## Checklist

1. Closes  #583

2. How was this tested:
- Ensured that `curl` embedded in the `admintools` support `--head` arg properly
- Tested locally

3. Any docs updates needed?
No


robholland:
The request isn't to check existence, it's trigger creation of an index based on the template we just added on the line before. I imagine the issue is that during an upgrade we get a 400 error because the index already exists. Changing it to run `head` first and then `PUT` if that returns 404 would probably work.

VLZZZ:
@robholland You're right, I didn’t notice at all that the template is first added and then an index is created from the template.

I'll now update my PR to something like 
`curl --head --fail xxx || template && PUT index` as you proposed

charts/temporal/templates/server-job.yaml
```diff
@@ -77,7 +77,8 @@ spec:
             {{- else if eq $driver "elasticsearch" }}
           command: ['sh', '-c']
           args:
-            - 'curl -X PUT --fail --user "$ES_USER:$ES_PWD" $ES_SCHEME://$ES_HOST:$ES_PORT/_template/temporal_visibility_v1_template -H "Content-Type: application/json" --data-binary "@schema/elasticsearch/visibility/index_template_$ES_VERSION.json" 2>&1 &&
+            - 'curl --head --fail --user "$ES_USER:$ES_PWD" $ES_SCHEME://$ES_HOST:$ES_PORT/$ES_VISIBILITY_INDEX 2>&1 ||
+              curl -X PUT --fail --user "$ES_USER:$ES_PWD" $ES_SCHEME://$ES_HOST:$ES_PORT/_template/temporal_visibility_v1_template -H "Content-Type: application/json" --data-binary "@schema/elasticsearch/visibility/index_template_$ES_VERSION.json" 2>&1 &&
```
robholland, line 81: We should always push the template. I don't believe that will fail if it already exists, it will just update. Just the index creation should be condtional.

VLZZZ: Thank you.
Makes sense, updated.

VLZZZ:
Hey, @robholland just checkin'
Do I need to rebase on master or we're good to proceed?
Also I see that `lint-test` is failing but as far as I see it's not related to my changes?

robholland:
Yes please rebase on master to get the CI fix, the error is indeed not your code.

robholland (approved):

PR #589, opened 2024-10-14
Print annotation and label value as YAML to support array string values

mark-hofmeijer:
## What was changed
This PR adds support for array strings as annotations and labels.

## Why?
Annotations and labels with array string as value (e.g. `ad.datadoghq.com/temporal-web.logs: '[{"source":"temporal","service":"temporal-web"}]'`) results in an error, both on installation and upgrade.

Error message: 
`Error: UPGRADE FAILED: cannot patch "temporal-admintools" with kind Deployment:  "" is invalid: patch: Invalid value: "{...}": json: cannot unmarshal array into Go struct field ObjectMeta.spec.template.metadata.annotations of type string`

## Checklist

1. How was this tested:
Automated.

Or manual via:
`helm install -f values.yaml --set web.podAnnotations.arrayannotation='[{"test":"true"}]' debug . --dry-run --debug`

2. Any docs updates needed?
No.

### DIFF
--- a/charts/temporal/templates/_helpers.tpl
+++ b/charts/temporal/templates/_helpers.tpl
@@ -91,11 +91,11 @@ Create the annotations for all resources
 {{- end -}}
 {{- $resourceAnnotations := merge $scopeAnnotations $componentAnnotations -}}
 {{- range $annotation_name, $annotation_value := $resourceAnnotations }}
-{{ $annotation_name }}: {{ $annotation_value }}
+{{ $annotation_name }}: {{ $annotation_value | toYaml }}
 {{- end -}}
 {{- end -}}
 {{- range $annotation_name, $annotation_value := $global.Values.additionalAnnotations }}
-{{ $annotation_name }}: {{ $annotation_value }}
+{{ $annotation_name }}: {{ $annotation_value | toYaml }}
 {{- end -}}
 {{- end -}}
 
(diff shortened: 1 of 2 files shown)


mark-hofmeijer:
@robholland Can you take a look at this one too? Although the latests PRs add support for "normal" string/integer/boolean values, it does not work for strings with arrays in it. This PR fixes this (and is eventually what I need :-)). Thanks!

stevengubler:
+1. We ran into this issue as well on the latest releases. Version `0.48.0` still respects structured data as annotations, so we've pinned the helm release to it as a temporary fix.

mark-hofmeijer:
@robholland Do you have time to give your review on this PR? Many thanks!

robholland:
`toYaml` doesn't seem right to me. The values have to land as strings. Shouldn't it just be ` | quote`?

robholland:
Put another way, the fact it happens to be an array when interpreted by whatever reads that annotation isn't really relevant. At the manifest YAML level all annotations and labels must be a string.

mark-hofmeijer:
@robholland Thank you for your feedback. I was not aware of this requirement, but it works with quoted values as well (and tests succeeds). 

Have looked at some other charts like those from [Datadog](https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/cluster-agent-deployment.yaml#L68) and [Istiod](https://github.com/istio/istio/blob/ba7ad009b737bb5361186f76bd4f7236111d03bd/manifests/charts/istio-control/istio-discovery/templates/deployment.yaml#L62) to see how they approach it and both charts were using toYaml so thats why I ended up with that.

robholland (approved):

PR #602, opened 2024-10-26
Add support for internal-frontend

dcaputo-harmoni:
## What was changed
Adds support for internal-frontend which has been available since v1.20

## Why?
Because it is necessary for certain auth configurations to allow internal access, also addresses #560

### DIFF
--- a/charts/temporal/templates/admintools-deployment.yaml
+++ b/charts/temporal/templates/admintools-deployment.yaml
@@ -33,13 +33,13 @@ spec:
           env:
             # TEMPORAL_CLI_ADDRESS is deprecated, use TEMPORAL_ADDRESS instead
             - name: TEMPORAL_CLI_ADDRESS
-              {{- if .Values.server.internalFrontend.enabled }}
+              {{- if and (hasKey .Values.server "internalFrontend") .Values.server.internalFrontend.enabled }}
               value: {{ include "temporal.fullname" $ }}-internal-frontend:{{ .Values.server.internalFrontend.service.port }}
               {{- else }}
               value: {{ include "temporal.fullname" $ }}-frontend:{{ .Values.server.frontend.service.port }}
               {{- end }}
             - name: TEMPORAL_ADDRESS
-              {{- if .Values.server.internalFrontend.enabled }}
+              {{- if and (hasKey .Values.server "internalFrontend") .Values.server.internalFrontend.enabled }}
               value: {{ include "temporal.fullname" $ }}-internal-frontend:{{ .Values.server.internalFrontend.service.port }}
               {{- else }}
               value: {{ include "temporal.fullname" $ }}-frontend:{{ .Values.server.frontend.service.port }}
--- a/charts/temporal/templates/admintools-deployment.yaml
+++ b/charts/temporal/templates/admintools-deployment.yaml
@@ -33,14 +33,14 @@ spec:
           env:
             # TEMPORAL_CLI_ADDRESS is deprecated, use TEMPORAL_ADDRESS instead
             - name: TEMPORAL_CLI_ADDRESS
-              {{- if index $.Values.server "internal-frontend" "enabled" }}
-              value: {{ include "temporal.fullname" $ }}-internal-frontend:{{ index $.Values.server "internal-frontend" "service" "port" }}
+              {{- if .Values.server.internalFrontend.enabled }}
+              value: {{ include "temporal.fullname" $ }}-internal-frontend:{{ .Values.server.internalFrontend.service.port }}
               {{- else }}
               value: {{ include "temporal.fullname" $ }}-frontend:{{ .Values.server.frontend.service.port }}
               {{- end }}
             - name: TEMPORAL_ADDRESS
-              {{- if index $.Values.server "internal-frontend" "enabled" }}
-              value: {{ include "temporal.fullname" $ }}-internal-frontend:{{ index $.Values.server "internal-frontend" "service" "port" }}
+              {{- if .Values.server.internalFrontend.enabled }}
+              value: {{ include "temporal.fullname" $ }}-internal-frontend:{{ .Values.server.internalFrontend.service.port }}
               {{- else }}
               value: {{ include "temporal.fullname" $ }}-frontend:{{ .Values.server.frontend.service.port }}
               {{- end }}
--- a/charts/temporal/templates/admintools-deployment.yaml
+++ b/charts/temporal/templates/admintools-deployment.yaml
@@ -33,9 +33,17 @@ spec:
           env:
             # TEMPORAL_CLI_ADDRESS is deprecated, use TEMPORAL_ADDRESS instead
             - name: TEMPORAL_CLI_ADDRESS
+              {{- if index $.Values.server "internal-frontend" "enabled" }}
+              value: {{ include "temporal.fullname" $ }}-internal-frontend:{{ index $.Values.server "internal-frontend" "service" "port" }}
+              {{- else }}
               value: {{ include "temporal.fullname" $ }}-frontend:{{ .Values.server.frontend.service.port }}
+              {{- end }}
             - name: TEMPORAL_ADDRESS
+              {{- if index $.Values.server "internal-frontend" "enabled" }}
+              value: {{ include "temporal.fullname" $ }}-internal-frontend:{{ index $.Values.server "internal-frontend" "service" "port" }}
+              {{- else }}
               value: {{ include "temporal.fullname" $ }}-frontend:{{ .Values.server.frontend.service.port }}
+              {{- end }}
           {{- if .Values.admintools.additionalEnv }}
           {{- toYaml .Values.admintools.additionalEnv | nindent 12 }}
           {{- end }}
--- a/charts/temporal/templates/server-configmap.yaml
+++ b/charts/temporal/templates/server-configmap.yaml
@@ -119,12 +119,12 @@ data:
           membershipPort: {{ $server.frontend.service.membershipPort }}
           bindOnIP: "0.0.0.0"
 
-      {{- if index $.Values.server "internal-frontend" "enabled" }}
+      {{- if .Values.server.internalFrontend.enabled }}
       internal-frontend:
         rpc:
-          grpcPort: {{ index $server "internal-frontend" "service" "port" }}
-          httpPort: {{ index $server "internal-frontend" "service" "httpPort" }}
-          membershipPort: {{ index $server "internal-frontend" "service" "membershipPort" }}
+          grpcPort: {{ $server.internalFrontend.service.port }}
+          httpPort: {{ $server.internalFrontend.service.httpPort }}
+          membershipPort: {{ $server.internalFrontend.service.membershipPort }}
           bindOnIP: "0.0.0.0"
       {{- end }}
 
@@ -182,7 +182,7 @@ data:
       {{- toYaml . | nindent 6 }}
     {{- end }}
 
-    {{- if not (index $.Values.server "internal-frontend" "enabled") }}
+    {{- if not (.Values.server.internalFrontend.enabled) }}
     publicClient:
       hostPort: "{{ include "temporal.componentname" (list $ "frontend") }}:{{ $server.frontend.service.port }}"
     {{- end }}
--- a/charts/temporal/templates/server-configmap.yaml
+++ b/charts/temporal/templates/server-configmap.yaml
@@ -119,6 +119,15 @@ data:
           membershipPort: {{ $server.frontend.service.membershipPort }}
           bindOnIP: "0.0.0.0"
 
+      {{- if index $.Values.server "internal-frontend" "enabled" }}
+      internal-frontend:
+        rpc:
+          grpcPort: {{ index $server "internal-frontend" "service" "port" }}
+          httpPort: {{ index $server "internal-frontend" "service" "httpPort" }}
+          membershipPort: {{ index $server "internal-frontend" "service" "membershipPort" }}
+          bindOnIP: "0.0.0.0"
+      {{- end }}
+
       history:
         rpc:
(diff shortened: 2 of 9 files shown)


charts/temporal/values.yaml
```diff
@@ -242,6 +242,33 @@ server:
     containerSecurityContext: {}
     topologySpreadConstraints: []
     podDisruptionBudget: {}
+  internal-frontend:
```
adamko147, line 245: seems to me the helm chart is uses `camelCase` convention, would it ok if we renamed `internal-frontend` to `internalFrontend`? It would also makes the usage a bit easier, we could use `{{ if $.Values.server.internalFrontend.enabled }}` directly.

adamko147, line 245: also, to be backwards compatible, should we disable internal frontend by default?

dcaputo-harmoni: @adamko147 I had initially done this as `internalFrontend` but recall these identifiers are actually used to generate the kubernetes services and it doesn't allow uppercase letters - hence `internal-frontend` which corresponds to what the actual temporal feature is named.

dcaputo-harmoni: @adamko147 agreed, that was meant to be disabled by default, I'll make this update

adamko147: huh, thanks, did not realize that.

dleblanc:
This seems to contain all of the work (and more) that was done in https://github.com/temporalio/helm-charts/pull/571. I'm happy to close that one if we can get this over the line.

dcaputo-harmoni:
@adamko147 @dleblanc just pushed a commit to address disabling by default, let me know if there's anything else needed here

adamko147 (commented):
👍looks good to me. still needs 👀 from code owners...

charts/temporal/values.yaml
TheGeniesis: Just as a proposition: you can create a simple helper to use internalFrontend as a "yaml" part and internal-frontend in names when needed:

```
{{- define "serviceName" -}}
    {{- $service := index . 0 -}}
    {{- if eq $service "internalFrontend" }}
        {{- print "internal-frontend" }}
    {{- else }}
        {{- print $service }}
    {{- end }}
{{- end -}}
```

Usage:

```
{{ $service := include "serviceName" (list "frontend" "internalFrontend" "matching" "history" "worker") }}
```

charts/temporal/templates/admintools-deployment.yaml
```diff
@@ -33,9 +33,17 @@ spec:
           env:
             # TEMPORAL_CLI_ADDRESS is deprecated, use TEMPORAL_ADDRESS instead
             - name: TEMPORAL_CLI_ADDRESS
+              {{- if index $.Values.server "internal-frontend" "enabled" }}
```
TheGeniesis, line 36: @adamko147 Just a question about security:

If we use internal frontend here then everyone who has access to the admin-tools pod will be able to execute commands without any additional auth. 

Is this expected?

charts/temporal/templates/server-configmap.yaml
```diff
@@ -173,8 +182,10 @@ data:
       {{- toYaml . | nindent 6 }}
     {{- end }}
 
+    {{- if not (index $.Values.server "internal-frontend" "enabled") }}
```
TheGeniesis, line 185: Did you test it with authorization? 

When I was preparing similar changes I had to add condition to "authorization" + "publicClient" only for frontend. Otherwise internalFrontend will try to do auth and fail. Also, if we skip auth section for frontend then users will be able to access UI without permission.

I ended up with creating a loop to make separated configmap per service

```yaml
{{- range $originalService := (list "frontend" "internalFrontend" "history" "matching" "worker") }}
{{ $serviceValues := index $.Values.server $originalService }}
{{ $service := include "serviceName" (list $originalService) }}
...
      {{- if eq $service "frontend" }}
      {{- with $server.config.authorization }}
      authorization:
        {{- toYaml . | nindent 10 }}
      {{- end }}
      {{- end }}
...
      {{- if ne $service "frontend"}}
      internal-frontend:
        rpc:
            grpcPort: {{ $server.internalFrontend.service.port }}
            httpPort: {{ $server.internalFrontend.service.httpPort }}
            membershipPort: {{ $server.internalFrontend.service.membershipPort }}
            bindOnIP: "0.0.0.0"
      {{- end }}
```

TheGeniesis (commented):
Thx for the MR, I'm waiting for an official implementation of feature!

charts/temporal/templates/admintools-deployment.yaml
adamko147: well, that’s how we use it in our use case. Temporal is running in dedicated cluster/namespace and only frontend is exposed to outside world. access to the k8s cluster controlled via IAM, so unless you don’t have cluster access, you don’t have admin tools access.

dcaputo-harmoni: @adamko147 what you described matches our use case as well

charts/temporal/templates/server-configmap.yaml
dcaputo-harmoni: @TheGeniesis We have tested it with authorization on frontend (in our case using OIDC) which is working, but I haven't specifically tested auth for internal-frontend.  I was under the impression (possibly mistakenly?) that the purpose of internal-frontend was to bypass authorization for internal access, but perhaps there's a middle ground where it has a different level of auth configured.  If you have insights or ideas here would appreciate any contributions to this PR as it's a bit outside of my wheelhouse.

adamko147: @TheGeniesis I don't think you're mistaken about the internal frontend purpose. It was introduces at https://github.com/temporalio/temporal/pull/3706 

> The server worker needs to make connections to the frontend with effective "admin" access (to all namespaces). With pluggable claim mappers and authorizers, we can't control the logic in them, and requiring all users to set up mTLS with special claims is a large burden. Adding internal frontends as an alternative is easier and more reliable.

In our use case we have https://docs.temporal.io/self-hosted-guide/security#plugins custom authorizer/claim mapper for JWT tokens issued by our IDP and there is no way how other temporal component could get/refresh the token, so internal-frontend solves exactly this case

dleblanc: I've successfully tested this PR with authorization enabled. Currently we're enabling auth using the built-in authorizer and claim plugins, and it's working without issue.

I haven't leveraged the features that were failing when we didn't have the internal worker enabled (eg: scheduled jobs), but the system did come up without the errors and seemed to be working properly.

I agree with @adamko147 about the internal frontend's purpose.

tomwheeler: I've been trying to find time in between meetings to test this, so I appreciate that confirmation, @dleblanc.

charts/temporal/templates/admintools-deployment.yaml
TheGeniesis: My internal "deny all" rule is screaming that it will be a problem for some people who has stricter polices related to this topic. But if code owners are ok with that then I'm ok too.

charts/temporal/templates/server-configmap.yaml
TheGeniesis: Maybe I wasn't clear enough.

**About definitions**
As you mentioned - internal-frontend shouldn't have any authorization and should by used by internal pods (like worker).

Frontend should have authorization and should be used by users (direct calls/web).

**My auth scenario**
I use Azure EntraID to manage accesses with default Temporal Authorizer + mapper.

The EntraID was configured by following a [medium article](https://phongthaicao.medium.com/temporal-authentication-and-authorization-using-azure-ad-f940646b61e0). The only difference is in system namespace which in the newer version is `temporal-system` instead of `system`.

**The problem which I faced**
During tests services were failing, because internal-frontend had "authorization" part in the config. I had to disable it for the internal-frontend, so it stopped trying to do auth for internal services.

The Frontend needs the "authorization" part and publicClient (since it's public).

**This MR**
The current config uses the same configmap for "frontend" and "internal-frontend". So, it's not possible to have different setup for both (internal without authorization, frontend without internal-frontend).

**How I solved the issue in my repo**
I have separated configmaps for frontend and internal-frontend with enabled/disabled options described above


I see it's working for you. There might be one more factor which I'm not aware of and it was the reason why it didn't fully work for me. Unfortunately, I'm behind my working schedule and I don't have time to test deployments again with this MR, to confirm that I'm able to reproduce the same problem again :/

charts/temporal/values.yaml
```diff
@@ -242,6 +242,35 @@ server:
...
+    # Enable this to create internal-frontend
+    enabled: false
+    service:
+      # Evaluated as template
+      annotations: {}
+      type: ClusterIP
+      port: 7233
+      membershipPort: 6933
+      httpPort: 7243
```
stephen-harris, line 254: Should these ports be 6936 and 7236 by default to match https://github.com/temporalio/temporal/blob/main/docker/config_template.yaml#L283-L284

tomwheeler: Ports 6936 and 7236 also matches the suggestion in [the release notes for version 1.20](https://github.com/temporalio/temporal/releases/tag/v1.20.0) (which introduced the internal-frontend).

dcaputo-harmoni:
Just trying to understand what needs to be done to get this merged. One comment referenced changing port numbers (which is a quick fix) but seems that's all that's outstanding?  Could one of the code owners confirm?

tomwheeler:
> Just trying to understand what needs to be done to get this merged. One comment referenced changing port numbers (which is a quick fix) but seems that's all that's outstanding? Could one of the code owners confirm?

@dcaputo-harmoni, I sincerely apologize for the delay here. Testing this is a manual process, and between meetings and other projects, I haven't had a contiguous block of time to see that all the way through yet. I will do that this week. 

There are two changes I would like to see:

1. The port numbers used here deviate from those mentioned in the 1.20 release notes and used in our docker-compose template. I'd prefer if this PR could be updated to use those port numbers.
2. It would be great if you could follow [the suggestion from @TheGeniesis|https://github.com/temporalio/helm-charts/pull/602#discussion_r1820513667] so that it would follow the camel-case convention (`internalFrontend`) in our Helm Chart but use the hyphenated form for the service in Kubernetes itself.

tomwheeler (changes requested):
I finished setting this up and reviewing the code this weekend. 

As mentioned earlier, there are two changes I would like to see:

1. The port numbers used for the internal-frontend in this PR are different from those mentioned in the 1.20 release notes and used in our docker-compose template. I'd prefer if this PR could be updated to use those port numbers.
2. For consistency with other parts of the Helm Chart, the user-facing code should use camel case (i.e., `internalFrontend`), while still using the hyphenated form (i.e., `internal-frontend`) for the output. The approach described [here](https://github.com/temporalio/helm-charts/pull/602#discussion_r1820513667) should work for this.

Thank you for your contribution and I apologize for the delay in my review. I should be able to merge this in within a day of seeing the requested changes.

dcaputo-harmoni:
@tomwheeler Just committed the requested changes, take a look and let me know if we are good to go here - thanks!  And when you get a chance would also appreciate it if you could review the [other PR](https://github.com/temporalio/helm-charts/pull/601) I have open relating to including username in the persistence secrets for enhanced security.

tomwheeler:
> Just committed the requested changes, take a look and let me know if we are good to go here
@dcaputo-harmoni I tried it this morning, but encountered an error `parse error at (temporal/templates/server-service.yaml:36): unexpected {{end}}`. I have a long series of meetings starting soon, and likely won't have time to dig into it until this evening, but wanted to mention it now in case you forgot to push a commit or something.

dcaputo-harmoni:
@tomwheeler my bad I have some additional code here that didn't get committed - will do shortly.

dcaputo-harmoni:
@tomwheeler Just committed the additional changes, and also added some checks to make it backwards compatible for anyone with a values file that doesn't have internalFrontend

tomwheeler (approved):
Approved. Thanks again for the contribution, @dcaputo-harmoni!

I'll review your other PR by Monday at the latest, likely sooner.

alexku7:
Looks like the https://github.com/temporalio/helm-charts/pull/602 didn't resolve the issue completely

If I enable the JWT auth in the helm chart, it's applied on the frontend and on the internal-fronted as well :(

dcaputo-harmoni:
@alexku7 Thanks for flagging this.  Saw you submitted #618, just want to confirm if that has addressed the issue.

alexku7:
> @alexku7 Thanks for flagging this.  Saw you submitted #618, just want to confirm if that has addressed the issue.

Not exactly 
The fix has addressed another minor issue with the internal frontend service.

Unfortunately in order to support different auth schemes for the frontend and for the internal frontend requires much more complex work:(
Looks like we need to separate the configMaps.

I don't see any other choice to support an ability to have an internal frontend with mtls auth while the external frontend supports the jwt only ( without the mtls) or vice versa

dcaputo-harmoni:
@alexku7 got it, my understanding was that internal-frontend was intended to have no auth, but if you wanted it to support a separate / different auth method, that could be added with some additional complexity as you noted.  It's working as-is for our application, but perhaps there will be others in the community with interest to help develop this.

alexku7:
> @alexku7 got it, my understanding was that internal-frontend was intended to have no auth, but if you wanted it to support a separate / different auth method, that could be added with some additional complexity as you noted.  It's working as-is for our application, but perhaps there will be others in the community with interest to help develop this.

I think, the best solution should allow the flexibility to set different set of authentication schemes.
If somebody wants to remove the auth completely on the internal frontend he should be able to doit but he also should be able to set a different auth scheme.

In my experience, usually we use the mtls for intra pod (including the frontend) and the jwt tokens for external frontend clients

Issue #236, opened 2021-10-22
[Feature Request] Use a Kubernetes secret to protect the OIDC client secret


srikanthchelluri:
Ideally, when we configure authentication to Temporal Web, we shouldn't pass the plaintext client secret as a Helm value but instead pass the name of the Kubernetes Secret which contains the client secret.

Right now, all of `.Values.web.config` is simply just dumped into a ConfigMap [1] then mounted to the Deployment [2]. Temporal Web reads from this configuration file [3].

Some initial ideas:

- Maybe we should be able to create a Kubernetes Secret separately that contains the entire configuration file and pass its name as a Helm value. Then we can mount the Secret in the web Deployment.
- We could possibly also just put the client secret in the Kubernetes secret (as opposed to the entire configuration file) though I haven't thought through exactly how that'd look in terms of templates.

[1] https://github.com/temporalio/helm-charts/blob/master/templates/web-configmap.yaml
[2] https://github.com/temporalio/helm-charts/blob/master/templates/web-deployment.yaml
[3] https://github.com/temporalio/web/blob/master/server/auth/index.js

Davidonium:
I think this would be good to have as I've been trying to set up SSO for a fresh temporal installation using the chart.

IMO the easiest would be for the client_secret to be an env var that could be injected from a secret. That'd require changes in the web codebase though.

Davidonium:
After some digging I found out that the `temporalio/ui` docker image [dockerizes](https://github.com/temporalio/ui-server/blob/main/docker/start-ui-server.sh#L10) the configuration file before running. So the `{{ .Env.TEMPORAL_* }} syntax can be used in the helm chart to achieve passing the client secret as a proper env var injected from a k8s secret.

peter-c-larsson:
Thanks for that information @Davidonium . Will try that approach will waiting for this issue to get attention.

enricojonas:
@debugger24 I tried this but it's not working and found out that it's not released yet. Is it possible to release it? It improves security a lot.

Thank you

robholland:
This was implemented in https://github.com/temporalio/helm-charts/pull/603 which has since been released.

PR #603, opened 2024-10-27
Refactor deployment templates to support additional env vars from Secret

debugger24:
## What was changed
Added option to pass secret name which will be mouted as env variables in the contianer. 

## Why?
This will help in scenarios like SSO integration or other conifgurations.

## Checklist

1. Closes 
https://github.com/temporalio/helm-charts/issues/236

2. How was this tested:
Using helm render locally.

3. Any docs updates needed?
Added `Enable SSO in Temporal UI` section in README.md

### DIFF
--- a/charts/temporal/templates/admintools-deployment.yaml
+++ b/charts/temporal/templates/admintools-deployment.yaml
@@ -39,7 +39,7 @@ spec:
           {{- if .Values.admintools.additionalEnv }}
           {{- toYaml .Values.admintools.additionalEnv | nindent 12 }}
           {{- end }}
-          {{- if.Values.admintools.additionalEnvSecretName }}
+          {{- if .Values.admintools.additionalEnvSecretName }}
           envFrom:
             - secretRef:
                 name: {{ .Values.admintools.additionalEnvSecretName }}
--- a/charts/temporal/templates/admintools-deployment.yaml
+++ b/charts/temporal/templates/admintools-deployment.yaml
@@ -39,6 +39,11 @@ spec:
           {{- if .Values.admintools.additionalEnv }}
           {{- toYaml .Values.admintools.additionalEnv | nindent 12 }}
           {{- end }}
+          {{- if.Values.admintools.additionalEnvSecretName }}
+          envFrom:
+            - secretRef:
+                name: {{ .Values.admintools.additionalEnvSecretName }}
+          {{- end }}
           livenessProbe:
               exec:
                 command:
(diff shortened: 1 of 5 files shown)


charts/temporal/values.yaml
```diff
@@ -263,6 +263,7 @@ server:
     tolerations: []
     affinity: {}
     additionalEnv: []
+    additionalEnvSecretName: ""
```
robholland, line 266: What do you think about `additionalEnvFromSecret`?

robholland: The value would be the name.

debugger24: `additionalEnvSecretName` is more clear it needs secret name here.

charts/temporal/templates/server-deployment.yaml
```diff
@@ -88,6 +88,11 @@ spec:
           {{- if or $.Values.server.additionalEnv $serviceValues.additionalEnv }}
           {{- toYaml (default $.Values.server.additionalEnv $serviceValues.additionalEnv) | nindent 12 }}
           {{- end }}
+          {{- if.Values.server.additionalEnvSecretName }}
```
robholland, line 91:
```suggestion
          {{- if .Values.server.additionalEnvSecretName }}
```

charts/temporal/templates/admintools-deployment.yaml
```diff
@@ -39,6 +39,11 @@ spec:
           {{- if .Values.admintools.additionalEnv }}
           {{- toYaml .Values.admintools.additionalEnv | nindent 12 }}
           {{- end }}
+          {{- if.Values.admintools.additionalEnvSecretName }}
```
robholland, line 42:
```suggestion
          {{- if .Values.admintools.additionalEnvSecretName }}
```

charts/temporal/templates/web-deployment.yaml
```diff
@@ -36,6 +36,11 @@ spec:
           {{- if .Values.web.additionalEnv }}
           {{- toYaml .Values.web.additionalEnv | nindent 12 }}
           {{- end }}
+          {{- if.Values.web.additionalEnvSecretName }}
```
robholland, line 39:
```suggestion
          {{- if .Values.web.additionalEnvSecretName }}
```

robholland (changes requested):

robholland (approved):

PR #604, opened 2024-11-04
add support for namespace search-attributes

Dariusch:
## What was changed

I am adding support for adding search attributes while creating a namespace.

## Why?
We recently migrated to the temporal chart and missed the option to include the previously used config.

## Checklist

1. How was this tested:

values:
```yaml
server:
  config:
    namespaces:
      # Enable this to create namespaces
      create: true
      namespace:
        - name: default
          retention: 3d
          searchAttributes: []
        - name: non-default
          retention: 3d
          searchAttributes:
            - name: my_keyword
              type: Keyword
            - name: my_keyword2
              type: Keyword
```

output:
```yaml
        - name: create-default-namespace
          image: "temporalio/admin-tools:1.25.1-tctl-1.18.1-cli-1.1.0"
          imagePullPolicy: IfNotPresent
          command: ['/bin/sh','-c']
          args: ['temporal operator namespace describe -n default || temporal operator namespace create -n default --retention 3d']
          env:
            - name: TEMPORAL_ADDRESS
              value: "release-name-temporal-frontend.rmt.svc:7233"
        - name: create-non-default-namespace
          image: "temporalio/admin-tools:1.25.1-tctl-1.18.1-cli-1.1.0"
          imagePullPolicy: IfNotPresent
          command: ['/bin/sh','-c']
          args: ['temporal operator namespace describe -n non-default || temporal operator namespace create -n non-default --retention 3d']
          env:
            - name: TEMPORAL_ADDRESS
              value: "release-name-temporal-frontend.rmt.svc:7233"
        - name: create-non-default-search-attributes
          image: "temporalio/admin-tools:1.25.1-tctl-1.18.1-cli-1.1.0"
          imagePullPolicy: IfNotPresent
          command: ['/bin/sh','-c']
          args: ['temporal operator search-attribute create --namespace non-default --name my_keyword --type Keyword --name my_keyword2 --type Keyword']
          env:
            - name: TEMPORAL_ADDRESS
              value: "release-name-temporal-frontend.rmt.svc:7233"
      containers:
        - name: done
          image: "temporalio/admin-tools:1.25.1-tctl-1.18.1-cli-1.1.0"
          imagePullPolicy: IfNotPresent
          command: ['sh', '-c', 'echo "Store setup completed"']
```

2. Any docs updates needed?

No, its an optional opt-in.


robholland (changes requested):
Search attributes do not belong to namespaces on self-hosted Temporal.

I'm happy to include this feature, but it needs to be separated from any namespace related code.

Dariusch:
What do you mean?

Afaik according to [the docs](https://docs.temporal.io/self-hosted-guide/visibility#:~:text=With%20the%20Visibility%20store%2C%20you%20can%20use%20List%20Filters%20with%20Search%20Attributes%20to%20list%20and%20filter%20Workflow%20Executions%20that%20you%20want%20to%20review.)  they are namespace scoped.

I would be happy to change the PR if you could point me in the right direction.

TheGeniesis:
@Dariusch You are right. Commands to add custom attributes per namespace were introduced in v1.25 (we use it in our project). If you are using Postgres be aware, this functionality [is bugged](https://github.com/temporalio/temporal/issues/6664) for now.

charts/temporal/templates/server-job.yaml
```diff
@@ -158,6 +158,28 @@ spec:
...
+        {{- if $namespace.searchAttributes }}
+        - name: create-{{ $namespace.name }}-search-attributes
+          image: "{{ $.Values.admintools.image.repository }}:{{ $.Values.admintools.image.tag }}"
+          imagePullPolicy: {{ $.Values.admintools.image.pullPolicy }}
+          command: ['/bin/sh','-c']
+          args: ['temporal operator search-attribute create --namespace {{ $namespace.name }} {{- range $attr := $namespace.searchAttributes }} --name {{ $attr.name }} --type {{ $attr.type }}{{- end }}']
+          env:
+            - name: TEMPORAL_ADDRESS
+              value: "{{ include "temporal.fullname" $ }}-frontend.{{ $.Release.Namespace }}.svc:{{ $.Values.server.frontend.service.port }}"
+              {{- with $.Values.server.additionalVolumeMounts }}
```
robholland, line 170:
```suggestion
              {{- with $.Values.admintools.additionalVolumeMounts }}
```

robholland (changes requested):
My apologies, this has changed since I last used it. They are indeed scoped to namespaces. One small change needed though.

robholland:
Also now learned that Elasticsearch implementation does not have per namespace search attributes, only SQL does. How do you think best to handle that?

Dariusch:
@robholland I added a condition, comment, and tests only to render the search attributes for the SQL driver

robholland:
I think this needs to be inverted, and rather a list of search attributes with an optional namespace field per attribute. This allows search attributes to be created regardless of whether namespaces are being auto-created. It also lines up with search attributes being a top-level concept for the CLI (https://docs.temporal.io/cli/operator#search-attribute). An error should be thrown (https://helm.sh/docs/chart_template_guide/function_list/#fail) when an invalid combination is used (namespace is set for a search attribute, but we're configured to use Elasticsearch driver).

supercairos:
This would be awesome to have :)

PR #605, opened 2024-11-04
Adding hpa in temporal

synected:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/charts/temporal/templates/server-hpa.yaml
+++ b/charts/temporal/templates/server-hpa.yaml
@@ -1,0 +1,40 @@
+{{- if $.Values.server.enabled }}
+{{- range $service := (list "frontend" "history" "matching" "worker") }}
+{{- $serviceValues := index $.Values.server $service -}}
+{{- if ($serviceValues.autoscaling.enabled) }}
+apiVersion: autoscaling/v2
+kind: HorizontalPodAutoscaler
+metadata:
+  name: {{ include "temporal.componentname" (list $ $service) }}
+  labels:
+    app.kubernetes.io/name: {{ include "temporal.name" $ }}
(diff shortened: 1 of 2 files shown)


robholland:
Thanks for the submission. We don't recommend autoscaling the Temporal Server components, with the possible exception of Frontend. History and Matching services must rebalance themselves when pods are added or removed, so ideally they would be scaled to handle peak load rather than scaled up and down. If you would like to refactor this to only scale Frontend, I'm happy to review it.

cancinar:
Hey @robholland, I would really appreciate it if you could elaborate on it a little bit more? Will this rebalancing of the History and Matching service happen automatically, or will they need to be scaled manually?

robholland:
Re-balancing and scaling are separate concepts, so I'm not sure what you are asking. If you scale History or Matching services (manually or otherwise) they will re-balance load (based on shards and partitions, respectively) every time a pod is added or removed. The consequence of this is that if you scale automatically, you can cause re-balancing to happen often, which can be bad for performance (depending on frequency and cluster load). We recommend you scale History and Matching to support peak load and don't autoscale.

robholland:
I'm going to close this, but please re-open if you'd like to strip down to just hpa for the frontend service.

esn89:
To confirm my understanding is right:

For history, matching and worker, it is best to scale vertically and not horizontally, right?

@robholland

robholland:
No, horizontal scaling is good. I just don't recommend automating it.

esn89:
@robholland 
Would it be fair to say that, we gotta do some capacity planning and pick a set of replicas and more or less stick to it?

PR #610, opened 2024-11-19
Add some unit tests.

robholland:
## What was changed
Add some unit tests.

## Why?
These are more succinct and much faster than the terratest unit tests.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


tomwheeler (approved):
I attempted to run this last night, but it failed. I pulled again this morning and found that the update made in the most recent commit fixed it.

PR #618, opened 2024-12-26
Fix internalFrontend service selector labels

alexku7:
## What was changed
The frontend-internal service will point to the corresponding internal-frontend pods 

## Why?
When the internalFrontend is enabled its corresponding service points to the frontend pods instead of the internal-Frontend pods.

## Checklist
Tested locally


tomwheeler (approved):

tomwheeler:
Thank you very much for your contribution!

PR #619, opened 2024-12-26
[Bugfix]: namespace .Values pointing to nil error fixed

itsneel:
## What was changed
`$` is appended to make `.Values` accessible in the template

## Why?
Variable global access fixed while creating name spaces in the server via helm

## Checklist

1. How was this tested:
helm charts built as expected

2. Any docs updates needed?
no. its just an values access fix in the template. no need for documentation to be updated


itsneel:
closing this as already fixed in #617

PR #620, opened 2024-12-26
fix whitespace

tomwheeler:
## What was changed

One line in the `charts/temporal/templates/server-job.yaml` file inadvertently had tabs instead of spaces, due to a mistake I made when resolving a merge conflict related to PR #602.

## Why?

When resolving a merge conflict, my editor used tabs instead of spaces in the YAML file.


tomwheeler:
Merging this without a review, since it undoes a mistake and the holiday and timezone differences will delay a review.

PR #624, opened 2025-01-13
Move additionalInitContainers to bottom in server-job.yaml

ragutsky:
## What was changed

Moved `admintools.additionalInitContainers` to the bottom in server-job.

## Why?

Since additional init containers run first there isn't a way to add additional steps to server job after schema setup/update (for instance, add search attributes).

## Checklist

1. Closes - 

2. How was this tested:

- add value:
```
admintools:
  additionalInitContainers:
    - name: create-search-attributes
      image: temporalio/admin-tools:1.26.2
      imagePullPolicy: IfNotPresent
      command: ['/bin/sh','-c']
      args: ['temporal operator search-attribute create --namespace "my-namespace" --name="CustomId" --type="Keyword"']
```
- run command `helm template demo charts/temporal -s templates/server-job.yaml`
- check that added container is at the end

3. Any docs updates needed?

No

### DIFF
--- a/charts/temporal/templates/server-job.yaml
+++ b/charts/temporal/templates/server-job.yaml
@@ -24,9 +24,6 @@ spec:
       {{ include "temporal.serviceAccount" $ }}
       restartPolicy: OnFailure
       initContainers:
-        {{- if $.Values.admintools.additionalInitContainers }}
-          {{- toYaml $.Values.admintools.additionalInitContainers | nindent 8 }}
-        {{- end }}
         {{- if $.Values.cassandra.enabled }}
         - name: check-cassandra-service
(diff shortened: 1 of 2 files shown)


robholland:
I appreciate the requirement here, but existing users rely on the `additionalInitContainers` behaving as-is in order to create cloud proxy sidecars etc to enable database connectivity. I think we should change to using `preInitContainers` and `postInitContainers` in values.yaml. `additionalInitContainers` would be removed from the example values.yaml file, but be supported as a legacy option and treated as `postInitContainers` which is more inline with the expectations of the name, albeit not what this template currently does. This would allow for both your use case and existing ones.

Issue #591, opened 2024-10-15
[Feature Request] Web Deployment has no probes


JRBANCEL:
### Is your feature request related to a problem? Please describe.
The Web Deployment has no probes (startup/liveliness/readiness).

### Describe the solution you'd like
Add probes, like on the other components.

PR #625, opened 2025-01-14
Added liveliness probe to the web deployment.

wbmoeller:
## What was changed
Modified the web-deployment.yml to specify a liveliness probe on the web component monitoring the http port.
Note: didn't add startup or readiness since the other components don't have these either.

## Why?
Improved monitoring of the web component.

## Checklist

1. Closes https://github.com/temporalio/helm-charts/issues/591
(only if you don't care about the readiness and startup probes)

2. How was this tested:
Used 'kubectl describe' on the web container to verify changes. 
Decreased initialDelaySeconds with a bad tcpSocket port to verify error condition.
Fixed the port to verify there's a Liveliness section in the description and no errors.

3. Any docs updates needed?
No

### DIFF
--- a/charts/temporal/templates/web-deployment.yaml
+++ b/charts/temporal/templates/web-deployment.yaml
@@ -42,13 +42,9 @@ spec:
                 name: {{ .Values.web.additionalEnvSecretName }}
           {{- end }}
           livenessProbe:
-            failureThreshold: 3
             initialDelaySeconds: 10
-            periodSeconds: 10
-            successThreshold: 1
             tcpSocket:
               port: http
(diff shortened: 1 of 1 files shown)


robholland (changes requested):
Please remove the settings which are the same as the defaults, at least `periodSeconds`, `successThreshold` and `timeoutSeconds` from a quick look.

wbmoeller:
@robholland thanks for the feedback - the default settings are cleaned up and tested. failureThreshold was the only additional setting that could be removed.

robholland (approved):

PR #633, opened 2025-01-26
Add support of clusterIP in server service

vovcyan:
## What was changed

Added support of passing `clusterIP` value for `temporal-chart-frontend` and `temporal-chart-internal-frontend` services.
## Why?

There is no options to pass `clusterIP` reserved value or `None` for `temporal-chart-frontend`.
## Checklist

1. Closes 

2. How was this tested: Unit tests

### DIFF
--- a/charts/temporal/templates/server-service.yaml
+++ b/charts/temporal/templates/server-service.yaml
@@ -13,6 +13,9 @@ metadata:
   {{- end }}
 spec:
   type: {{ $serviceValues.service.type }}
+  {{- if hasKey $serviceValues.service "clusterIP" }}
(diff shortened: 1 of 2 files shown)


charts/temporal/templates/server-service.yaml
```diff
@@ -13,6 +13,9 @@ metadata:
   {{- end }}
 spec:
   type: {{ $serviceValues.service.type }}
+  {{- if hasKey $serviceValues.service "clusterIP" }}
```
robholland, line 16: Please switch this to `with` and use `.` in the block, rather than `if`.

robholland (changes requested):

Issue #338, opened 2022-12-07
Temporal services are not starting when deployed in istio enabled namespace


mmlk09:
### What are you really trying to do?

Trying to deploy temporal in istio enabled namespace along with PostgreSQL DB on a Kubernetes cluster.
Deployment is done using helm charts available at - https://github.com/temporalio/helm-charts

### Describe the bug

Post deployment history, matching and worker PODs are continuously restarting with CrashLoopBackOff with below mentioned logs. 

**Here is status of PODs:**
NAME                                       READY   STATUS             RESTARTS        AGE
temporaldb-7d7f9c67b8-qqxbb                2/2     Running            0               22m
temporaltest-admintools-7bf55d4b8f-t47m2   2/2     Running            0               16m
temporaltest-frontend-546c569b9b-g4gv4     2/2     Running            0               16m
temporaltest-history-8648bd7db9-xsnz4      1/2     CrashLoopBackOff   6 (74s ago)     16m
temporaltest-matching-bc9f69487-9vc95      1/2     CrashLoopBackOff   6 (52s ago)     16m
temporaltest-web-67d46c8748-zf4xn          2/2     Running            0               16m
temporaltest-worker-5c58f8cc58-c6jwd       1/2     Error              6 (4m32s ago)   16m

**LOGs:**
{"level":"info","ts":"2022-12-07T17:48:04.721Z","msg":"bootstrap hosts fetched","service":"worker","bootstrap-hostports":"10.4.0.125:6939,10.4.0.124:6933,10.4.0.126:6935,10.4.0.127:6934","logging-call-at":"rpMonitor.go:285"}
[...]

PR #639, opened 2025-01-29
Fix ports for service and enable istio support out-of-the-box

Hronom:
## What was changed

Fix ports for service and enable Istio support out-of-the-box.

## Checklist

1. Closes https://github.com/temporalio/helm-charts/issues/338

2. How was this tested:

Tested on cluster with Istio enabled and strict mTLS.

3. Any docs updates needed?

### DIFF
--- a/charts/temporal/templates/server-deployment.yaml
+++ b/charts/temporal/templates/server-deployment.yaml
@@ -102,12 +102,17 @@ spec:
             - secretRef:
                 name: {{ $.Values.server.additionalEnvSecretName }}
           {{- end }}
+          # For Istio service mesh - make sure ports are defined here and in the headless service, see:
+          # https://istio.io/latest/docs/ops/configuration/traffic-management/traffic-routing/#headless-services
           ports:
           {{- if ne $service "worker" }}
             - name: rpc
               containerPort: {{ $serviceValues.service.port }}
               protocol: TCP
           {{- end }}
+            - name: membership
+              containerPort: {{ $serviceValues.service.membershipPort }}
+              protocol: TCP
           {{- if or (eq $service "frontend") (eq $service "internal-frontend") }}
             - name: http
               containerPort: {{ $serviceValues.service.httpPort }}
--- a/charts/temporal/templates/server-service.yaml
+++ b/charts/temporal/templates/server-service.yaml
@@ -61,13 +61,27 @@ spec:
   type: ClusterIP
   clusterIP: None
   publishNotReadyAddresses: true
+  # For Istio service mesh - make sure all ports are defined here and in the deployment:
+  # https://istio.io/latest/docs/ops/configuration/traffic-management/traffic-routing/#headless-services
+  # Also for Istio - make sure to set the `appProtocol` property, see:
+  # https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection
+  # Note that only the monitoring port is used for discovery (by prometheus).
+  # The other ports are listed here solely to allow Istio to configure itself to intercept traffic.
+  # https://istio.io/latest/docs/ops/configuration/traffic-management/traffic-routing/#headless-services
   ports:
     - port: {{ $serviceValues.service.port }}
       targetPort: rpc
+      appProtocol: tcp
       protocol: TCP
       name: grpc-rpc
+    - port: {{ $serviceValues.service.membershipPort }}
+      targetPort: membership
+      appProtocol: tcp
+      protocol: TCP
+      name: grpc-membership
     - port: 9090
       targetPort: metrics
+      appProtocol: http
       protocol: TCP
       name: metrics
   selector:


Hronom:
Also I signed the CLA, not sure why the old comment is drawn

charts/temporal/templates/server-service.yaml
```diff
@@ -64,10 +64,17 @@ spec:
   ports:
     - port: {{ $serviceValues.service.port }}
       targetPort: rpc
+      appProtocol: tcp
       protocol: TCP
       name: grpc-rpc
+    - port: {{ $serviceValues.service.membershipPort }}
```
robholland, line 70: This port should not be wired up to a service, it's only used for direct connections between the pods.

robholland (changes requested):

charts/temporal/templates/server-service.yaml
Hronom: @robholland this change is for a `headless` service and not for regular service(that stays in this file upper). So `headless` service used for pod-to-pod communication https://stackoverflow.com/a/52713482. 

Moreover, without this change, Istio not work properly. This is main intention of this PR is to make temporal compatible with Istio setup that has strict mTLS enabled.

Hronom: Added comments inside a file to reflect why this change is needed. For future maintainability.

robholland: The headless service is not for pod to pod communication, it's only used for prometheus monitoring scraping. Temporal does not make use of any Kubernetes services internally.

robholland: Hmm. I don't want to increase confusion about what these services are used for, it already often causes confusion. It's a shame that Istio relies on this indirect mechanism rather than a Deployment annotation or similar. Please add a new value `istio: true|false` (defaulted `false`). The monitor port should always be in the headless service, all other ports should only be added if `istio` is true. It's fine to leave `appProtocol` on the metrics port unconditionally.

robholland: I think you can remove your comments once the ports are conditional, it should be self documenting enough that people can infer the reasons. Maybe include a one line comment above the `istio` value line explaining that it enables support for Istio service configuration.

robholland (changes requested):

charts/temporal/templates/server-service.yaml
Hronom: @robholland I would recommend to not introduce `istio` flag and make it like proposed in PR. Otherwise for those who use this helm chart - there will be confusion and expectation that there will be also added resources specific to the Istio like `kind: VirtualService` etc.

What PR's is really doing is making this helm chart friendly to be used in Kubernetes setups that use Istio out-of-the box.
In this PR we don’t introduce direct dependency on Istio.

Also from what I saw for helm charts that we use(we use many) for open-source tools - maintainers do not introduce special variable like `istio` to enable compatibility with Istio. They adjust helm charts to have this change as they are compatible with setup without Istio e.g. not change default behaviour.

Example: Mimir, Tempo(https://github.com/grafana/tempo/issues/4326). They also have cluster discovery e.g. pod-to-pod.

Please let me know if my thoughts changed your mind and make sense. 
And if not - I will introduce what you requested.

robholland: It's not behaviour changes I'm worried about, but more confusion about what our headless services are for. Mimir is not a good comparison because they do use k8s for discovery: https://github.com/grafana/mimir/blob/ce9463ed2d6c864b1bc7f327471f38f746d05fc5/operations/helm/charts/mimir-distributed/values.yaml#L359. We don't do that, but adding the ports here implies that we do.

Let me consult with the team and see what they think.

robholland: Ok, so unfortunately I'm not an expert in Istio, but do you think this on the deployments would work as an alternative:

```proxy.istio.io/config: |
  proxyMetadata:
    ISTIO_META_INTERCEPTION_MODE: REDIRECT
```

As far as I understand it the addition of the headless service port is just to tell Istio to create a listener for `<POD_IP>:<PORT>`, I believe the above would have the effect of creating a listener for `<POD_IP:*>`. As we don't use the k8s for discovery we don't need the DNS features that the headless service port provides.

Doing it this way would avoid creating confusing service ports that aren't needed and also makes it very clear what is Istio specific.

Is that something you could test? Did I misunderstand something about requirements?

Hronom: Let me try to create a test stand that I can share.

robholland: I've done some more research into this, I think I'm wrong and the redirection annotation will not help here. Let's go ahead with your changes and I'll continue to think about ways we might reduce confusion.

charts/temporal/templates/server-deployment.yaml
```diff
@@ -102,12 +102,19 @@ spec:
             - secretRef:
                 name: {{ $.Values.server.additionalEnvSecretName }}
           {{- end }}
+          # For Istio service mesh - make sure ports a properly defined to be used in headless service, see:
```
robholland, line 105:
```suggestion
          # For Istio service mesh - make sure ports are defined here and in the headless service, see:
```

charts/temporal/templates/server-deployment.yaml
```diff
@@ -102,12 +102,19 @@ spec:
...
+          # https://istio.io/latest/docs/ops/configuration/traffic-management/traffic-routing/#headless-services
+          # So when you add here port involved into pod-to-pod communication - make sure that you add this port
+          # to the headless service too.
```
robholland, line 108:
```suggestion
```

charts/temporal/templates/server-service.yaml
```diff
@@ -61,13 +61,26 @@ spec:
   type: ClusterIP
   clusterIP: None
   publishNotReadyAddresses: true
+  # For Istio service mesh - make sure ports a properly defined in headless service, see:
+  # https://istio.io/latest/docs/ops/configuration/traffic-management/traffic-routing/#headless-services
+  # Make sure here that ports involved in pod-to-pod communication also properly defined inside service deployment.
+  #
+  # Also for Istio - make sure that `appProtocol` property - properly defined, see:
+  # https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection
```
robholland, line 69:
```suggestion
  # For Istio service mesh - make sure all ports are defined here and in the deployment:
  # https://istio.io/latest/docs/ops/configuration/traffic-management/traffic-routing/#headless-services
  # Also for Istio - make sure to set the `appProtocol` property, see:
  # https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection
```

charts/temporal/templates/server-service.yaml
```diff
@@ -61,13 +61,26 @@ spec:
...
   ports:
```
robholland, line 70:
```suggestion
  # Note that only the monitoring port is used for discovery (by prometheus).
  # The other ports are listed here solely to allow Istio to configure itself to intercept traffic.
  # https://istio.io/latest/docs/ops/configuration/traffic-management/traffic-routing/#headless-services
  ports:
```

robholland (approved):
Please tweak the docs a little as per suggestions, but otherwise happy to merge. Thanks for your patience while exploring alternatives.

Hronom:
@robholland applied suggested changes. Please trigger that job that is currently blocking the merge. Thanks!

Hronom:
@robholland one more time, I merged suggestions. Seems like only you can merge it. Let me know if I missed anything

robholland:
Thanks for your work and patience :)

PR #641, opened 2025-02-05
[hem-chart] add extra parameters to temporal-sql-tool

luismacosta:
Hello @robholland 

I would like to create my own job for db bootstrap, as I am using RDS postgresql/aurora and need to pass extra parameters in temporal-sql-tool, such as:

```
--tls --tls-ca-file
--ca search_path
--pl postgres12 (due to breaking change on v1.20.0 - https://github.com/temporalio/temporal/releases/tag/v1.20.0)
```

## What was changed
temporal_sql_tool_paramaters added


luismacosta:
temporal_sql_tool_paramaters added

robholland:
So looking into this more, I understand that your reason for wanting this is solely to be able to set `--pl postgres12` while still having the SQL driver set to `postgres` in order to allow `sql-tool` to run with a configuration different to that of the server while you perform an upgrade?

luismacosta:
> So looking into this more, I understand that your reason for wanting this is solely to be able to set `--pl postgres12` while still having the SQL driver set to `postgres` in order to allow `sql-tool` to run with a configuration different to that of the server while you perform an upgrade?

No, I also need to define: 

--ca search_path: specify the database schema
--tls --tls-ca-file: The rds.force_ssl parameter default value is 1 (on) for RDS for [PostgreSQL version 15 and later](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/PostgreSQL.Concepts.General.SSL.html). Since the docker image does not container the global pem, I need to add the global [certificate bundle](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/UsingWithRDS.SSL.html)

robholland:
Those are all available via admintools-env: https://github.com/temporalio/helm-charts/blob/main/charts/temporal/templates/_admintools-env.yaml so I don't think we need to add any new feature for that?

robholland:
If the issue is solely having to do this for the commands while leaving the server config as it was for an upgrade cycle, I don't think we should add the complexity to the chart for this use case. I suggest just running the commands manually before upgrading via helm.

luismacosta:
> Those are all available via admintools-env: https://github.com/temporalio/helm-charts/blob/main/charts/temporal/templates/_admintools-env.yaml so I don't think we need to add any new feature for that?

Ok, got it, I just need to define _caFile_

> ```
> https://github.com/temporalio/temporal/releases/tag/v1.20.0)
> ```

luismacosta:
> If the issue is solely having to do this for the commands while leaving the server config as it was for an upgrade cycle, I don't think we should add the complexity to the chart for this use case. I suggest just running the commands manually before upgrading via helm.

Ok, as mentioned in the release notes for [v1.20.0](https://github.com/temporalio/temporal/releases/tag/v1.20.0), run manually temporal-sql-tool with --pl postgres12.

Thanks a loot.
I will close this PR.

PR #642, opened 2025-02-05
Add AdditionalEnvs to Namespace Job

### LINKED ISSUE #640: [Feature Request] Ability to add Additional Envs to CreateNamespace Job
Having the ability to add additional environment variables for the create namespace would be helpful - particularly when arguments related to TLS need to passed in to the CLI/commands.

Adding `additionalEnv: []` either at `schema` or `server.config.namespaces` then injecting it into the namespace job with
```yaml
env:
...
{{- if $.Values.server.config.namespaces.additionalEnv }} 
{{- toYaml $.Values.server.config.namespaces.additionalEnv | nindent 12 }}
{{- end }}    
```

lsu-tc:
## What was changed
Added ability to specify additional environment variables for the namespace job in `server-job.yaml`

## Why?
Allows TLS arguments to be passed to commands used by Temporal CLI

## Checklist
1. Closes #640 

2. How was this tested:
 - set `server.config.namespaces.create` to true
 - add the following to `env` in the `create-{{ $namespace.name }}-namespace` job, at line 153
 ```yaml
{{- with $.Values.admintools.additionalEnv }} 
  {{- toYaml . | nindent 12 }}
{{- end }} 
```
 - add any envs under `admintools.additionalEnv`
 - run `helm template demo charts/temporal -s templates/server-job.yaml`
 - verify that the envs under the create namespace jobs reflect the envs added
 
3. Any docs updates needed?
None

### DIFF
--- a/charts/temporal/templates/server-job.yaml
+++ b/charts/temporal/templates/server-job.yaml
@@ -150,8 +150,8 @@ spec:
               {{- else }}
               value: "{{ include "temporal.fullname" $ }}-frontend.{{ $.Release.Namespace }}.svc:{{ $.Values.server.frontend.service.port }}"
               {{- end }}
-              {{- if $.Values.admintools.additionalEnv }} 
-                {{- toYaml $.Values.admintools.additionalEnv | nindent 12 }}
+              {{- with $.Values.admintools.additionalEnv }}
+                {{- toYaml . | nindent 12 }}
               {{- end }} 
               {{- with $.Values.admintools.additionalVolumeMounts }}
           volumeMounts:
--- a/charts/temporal/templates/server-job.yaml
+++ b/charts/temporal/templates/server-job.yaml
@@ -150,6 +150,9 @@ spec:
               {{- else }}
               value: "{{ include "temporal.fullname" $ }}-frontend.{{ $.Release.Namespace }}.svc:{{ $.Values.server.frontend.service.port }}"
               {{- end }}
+              {{- if $.Values.admintools.additionalEnv }} 
+                {{- toYaml $.Values.admintools.additionalEnv | nindent 12 }}
+              {{- end }} 
               {{- with $.Values.admintools.additionalVolumeMounts }}
           volumeMounts:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by lsu-tc** [comment]
Only the create namespace job needs additional envs because it is using the Temporal Operator CLI. In the case that mTLS for internode is configured, the CLI expects TLS related arguments which can be provided from the additional envs. 

If not provided, the job will show an error relating to tls: `msg="failed reaching server: last connection error: connection error: desc = \"error reading server preface: EOF\"`

**Comment by robholland** [comment]
I think the correct fix here is to add the missing include for the `admintools-env` helper to that pod, like the other admintools based pods have. Then also add `admintools.additionalEnv` to that helper, which it currently lacks. This makes the server job more consistent, and also allows the admintools deployment to have the correct mTLS environment so that `temporal operator` will work out of the box. Whichever env vars you need for internode mTLS would then go in the `admintools.additionalEnv` section in values.yaml.

File: charts/temporal/values.yaml [hunk 1]
```diff
@@ -202,6 +202,7 @@ server:
       namespace:
         - name: default
           retention: 3d
+      additionalEnv: []
```
robholland, line 205: This change can be reverted now.

robholland (changes requested):

charts/temporal/templates/server-job.yaml
```diff
@@ -150,6 +150,9 @@ spec:
               {{- else }}
               value: "{{ include "temporal.fullname" $ }}-frontend.{{ $.Release.Namespace }}.svc:{{ $.Values.server.frontend.service.port }}"
               {{- end }}
+              {{- if $.Values.admintools.additionalEnv }}
```
robholland, line 153: Please switch to `with`.

robholland (changes requested):
Happy to approve and merge with a small tweak.

charts/temporal/templates/server-job.yaml
lsu-tc: Switched!

PR #643, opened 2025-02-07
fix(elasticsearch): make auth parameters optional in health check

### LINKED ISSUE #621: [Bug] Temporal pods infinite in init state when Elasticsearch has ho username/password auth configured
### Describe the bug

After update Temporal to the latest release all pods hung in init state at running `check-elasticsearch-index` container.
Logs have entries:
`waiting for elasticsearch index to become ready`

After investigation it found that Elasticsearch return 403 when configured to have no username/password for requests having auth header.

Seems the issue appeared at this commit:
https://github.com/temporalio/helm-charts/commit/6d50f5d28c053e987c0486f2ecc2d56a84a1d689#diff-b8142966d328045abc62158c9a0c7b8492a97eb135757aea4a6941d4a80edb9eR58

Previously helm ommited curl `--user` argument when no username/password was configured; now the resulted curl command
`curl -i --fail --user ":" https://elasticsearch:443/temporal_visibility_v1_dev` return 403 
 
### Minimal Reproduction

Configure Elasticsearch as visibility storage without username/password authentication.
Deploy temporal helm chart v0.54.0

### Temporary workaround
Patch the template to omit entire `--user "$ES_USER:$ES_PWD"` argument in curl command

NunoMaga:
ix: Only add --user flag when both ES_USER and ES_PWD env vars are set. Previously the curl command always included auth headers even when not configured, causing Elasticsearch to reject the request.  Resolves issue where pods would hang in init state when Elasticsearch auth was not configured.

## What was changed
Makes the --user parameter conditional in the Elasticsearch health check command,
only including authentication when both ES_USER and ES_PWD environment variables
are set.

## Why?
When Elasticsearch is configured without authentication, the init container was sending empty auth credentials (--user ":") which caused 403 errors. This change makes the --user flag conditional on having both username and password set.  Bug: Init containers would hang when checking Elasticsearch readiness due to 403 responses from sending empty auth credentials.  

1. Closes  #621 

2. How was this tested:

### DIFF
--- a/charts/temporal/templates/server-deployment.yaml
+++ b/charts/temporal/templates/server-deployment.yaml
@@ -60,7 +60,7 @@ spec:
         - name: check-elasticsearch-index
           image: "{{ $.Values.admintools.image.repository }}:{{ $.Values.admintools.image.tag }}"
           imagePullPolicy: {{ $.Values.admintools.image.pullPolicy }}
-          command: ['sh', '-c', 'until curl --silent --fail --user "$ES_USER:$ES_PWD" $ES_SCHEME://$ES_HOST:$ES_PORT/$ES_VISIBILITY_INDEX 2>&1 > /dev/null; do echo waiting for elasticsearch index to become ready; sleep 1; done;']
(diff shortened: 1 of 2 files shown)


robholland:
Hey @NunoMaga, do you need any help here?

Confushion:
Any progress on merging this bugfix?
I could really use this fix in my deployments...

NunoMaga:
totally forgot this pr and have completely lost context. unsure how I tested it

robholland:
We no longer set user/password unless they are present.

PR #646, opened 2025-02-14
[helm-charts][web] add additionalInitContainers

luismacosta:
[helm-charts][web] add additionalInitContainers

## What was changed
_additionalInitContainers_ added to web-deployment.yaml

### DIFF
--- a/charts/temporal/templates/web-deployment.yaml
+++ b/charts/temporal/templates/web-deployment.yaml
@@ -23,7 +23,9 @@ spec:
     spec:
       initContainers:
         {{- if $.Values.web.initContainers }}
-        {{- toYaml $.Values.server.initContainers | nindent 8}}
+        {{- with .Values.web.initContainers }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
         {{- end }}
       {{ include "temporal.serviceAccount" $ }}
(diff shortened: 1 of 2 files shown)


luismacosta (commented):
change this

charts/temporal/templates/web-deployment.yaml
```diff
@@ -21,6 +21,12 @@ spec:
       labels:
         {{- include "temporal.resourceLabels" (list $ "web" "pod") | nindent 8 }}
     spec:
+      {{- if .Values.web.initContainers }}
```
robholland, line 24: You don't need to guard with `if`, `with` correctly handles the value not being present.

charts/temporal/values.yaml
```diff
@@ -373,6 +373,7 @@ admintools:
   securityContext: {}
   podDisruptionBudget: {}
 web:
+  # initContainers: []
```
robholland, line 376: Please make this `additionalInitContainers` to match the API for this functionality for other deployments.

charts/temporal/templates/web-deployment.yaml
luismacosta: done

robholland (approved):

Issue #444, opened 2023-11-21
[Feature Request] sidecar container for schema creation & update jobs


RonaldGalea:
### Is your feature request related to a problem? Please describe.

It is currently possible to connect to GCP Cloud SQL via the sidecar proxy - but the schema creation & update jobs do not run this sidecar, meaning one has to manually do it.

### Describe the solution you'd like

If the schema jobs ran the sidecar, the process would be streamlined. Is there any reason not to do it that I'm missing?

Thanks.

jsok:
This is complicated by the fact that the schema setup is done via `initContainer`s, which IIUC run before `sideContainer`s start, i.e. the sql proxy won't be running yet.

enricojonas:
Facing the same issue. Basically we should have the possibility to have the sidecar setup for the scheme creation and also for the admin tools.

robholland:
You can use `admintools.additionalInitContainers` with `restartPolicy: Always` to achieve this now.

PR #649, opened 2025-02-18
Remove outdated sidecar support.

robholland:
## What was changed

Removed sidecarContainers value.

Swapped some `if` for `with` while I was working on the files to simplify them slightly.

## Why?
Recent Kubernetes versions have specific support for sidecar via the initContainers field.
https://kubernetes.io/docs/concepts/workloads/pods/sidecar-containers/

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


robholland:
Also gives an example of how we support https://github.com/temporalio/helm-charts/issues/444 now.

tomwheeler (approved):
Looks good to me. All tests pass and I was able to deploy a cluster using the code in this branch.

tomwheeler:
I have verified that `helm unittest` passes with Rob's latest commit.

PR #650, opened 2025-02-19
[frontend] define serviceName

luismacosta:
## What was changed
[frontend] define serviceName

## Why?
I'm using [ingress-nginx controller](https://kubernetes.github.io/ingress-nginx/user-guide/exposing-tcp-udp-services/), to expose multiple temporal clusters frontend services in different [tcp](https://github.com/kubernetes/ingress-nginx/blob/main/charts/ingress-nginx/values.yaml#L192) ports, using the same NLB. As such, I need to set the serviceName to a specific value.

e.g.
```yaml
tcp:
  7233: "temporal-test1/temporal-frontend-serviceName:7233"
  7234: "temporal-test2/temporal-frontend-serviceName:7233"
  7235: "temporal-test3/temporal-frontend-serviceName7233"
```

2. How was this tested:
```yaml
cd charts/temporal
helm template . --name-template temporal -n temporal -f values.yaml > test.yaml --debug
```

@robholland

### DIFF
--- a/charts/temporal/values.yaml
+++ b/charts/temporal/values.yaml
@@ -140,7 +140,7 @@ server:
         driver: "cassandra"
         cassandra:
           hosts: []
-          # port: 9042
+          port: 9042
           keyspace: "temporal"
           user: "user"
           password: "password"
@@ -170,7 +170,7 @@ server:
         driver: "cassandra"
         cassandra:
           hosts: []
-          # port: 9042
+          port: 9042
           keyspace: "temporal_visibility"
           user: "user"
           password: "password"
@@ -204,6 +204,7 @@ server:
           retention: 3d
   frontend:
     service:
+      # name: temporal-frontend-name
       # Evaluated as template
       annotations: {}
       type: ClusterIP
@@ -211,7 +212,7 @@ server:
       membershipPort: 6933
       httpPort: 7243
     ingress:
-      enabled: false
+      enabled: true
       # className:
       annotations: {}
       # kubernetes.io/ingress.class: traefik
--- a/charts/temporal/values.yaml
+++ b/charts/temporal/values.yaml
@@ -204,7 +204,7 @@ server:
           retention: 3d
   frontend:
     service:
-      # name: temporal-frontend-name
+      # name: temporal-frontend-serviceName
       # Evaluated as template
       annotations: {}
       type: ClusterIP
--- a/charts/temporal/values.yaml
+++ b/charts/temporal/values.yaml
@@ -170,7 +170,7 @@ server:
         driver: "cassandra"
         cassandra:
           hosts: []
-          port: 9042
+          # port: 9042
           keyspace: "temporal_visibility"
           user: "user"
           password: "password"
--- a/charts/temporal/values.yaml
+++ b/charts/temporal/values.yaml
@@ -140,7 +140,7 @@ server:
         driver: "cassandra"
         cassandra:
           hosts: []
-          port: 9042
+          # port: 9042
           keyspace: "temporal"
           user: "user"
           password: "password"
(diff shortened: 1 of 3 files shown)


charts/temporal/values.yaml
```diff
@@ -140,7 +140,7 @@ server:
         driver: "cassandra"
         cassandra:
           hosts: []
-          # port: 9042
+          port: 9042
```
luismacosta, line 143:
```suggestion
          # port: 9042
```

charts/temporal/values.yaml
```diff
@@ -170,7 +170,7 @@ server:
...
```
luismacosta, line 173:
```suggestion
          # port: 9042
```

charts/temporal/values.yaml
```diff
@@ -204,14 +204,15 @@ server:
...
+      # name: temporal-frontend-name
       # Evaluated as template
       annotations: {}
       type: ClusterIP
       port: 7233
       membershipPort: 6933
       httpPort: 7243
     ingress:
-      enabled: false
+      enabled: true
```
luismacosta, line 215:
```suggestion
      enabled: false
```

charts/temporal/values.yaml
```diff
@@ -204,6 +204,7 @@ server:
           retention: 3d
   frontend:
     service:
+      # name: temporal-frontend-name
```
luismacosta, line 207:
```suggestion
      # name: temporal-frontend-serviceName
```

robholland:
Sorry, I don't understand the use case. These clusters are either in different namespaces and/or have different releases names. So why do you need to change the frontend service name?

luismacosta:
> Sorry, I don't understand the use case. These clusters are either in different namespaces and/or have different releases names. So why do you need to change the frontend service name?

Once I'm deploying the clusters via ArgoCD, the serviceName will contain a random ID e.g. _temporal-test-1-5aabedce12a6d5f-frontend_ . I need a static name for the service, to be used in ingress-nginx controller. e.g. _temporal-test-1-frontend_

robholland:
I believe you can set the release name to get consistent naming here: https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#helm-release-name

luismacosta:
> I believe you can set the release name to get consistent naming here: https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#helm-release-name

ok, I will have a look. but why not give the possibility in the helm chart to do it?

robholland:
Because that would add a non-standard feature to our chart when it seems there is a common way to solve this issue that doesn't require extra code, testing or documentation on our part.

robholland:
Why would this impact other applications?

luismacosta:
> Why would this impact other applications?

I've one appSet looking into the path where are all apps. I don't want to touch the remaining apps. But I can keep the approach that is implemented at the moment, I've another service.yaml template for temporal clusters where serviceName is defined as a static value.

robholland:
I still don't fully understand. But if you don't want to do it at Argo level, you can also set the `nameOverride` or `fullnameOverride` values: https://github.com/temporalio/helm-charts/blob/main/charts/temporal/values.yaml#L1

Those should also allow you to control the naming of services across the entire Temporal install.

luismacosta:
> I still don't fully understand. But if you don't want to do it at Argo level, you can also set the `nameOverride` or `fullnameOverride` values: https://github.com/temporalio/helm-charts/blob/main/charts/temporal/values.yaml#L1
> 
> Those should also allow you to control the naming of services across the entire Temporal install.

OK, so nameOverride will override the servivceName? https://github.com/temporalio/helm-charts/blob/main/charts/temporal/templates/server-service.yaml#L8

robholland:
I'm not sure exactly where the random character suffix is coming from in your case, but either `nameOverride` or `fullnameOverride` will allow you to make the component names consistent, yes.

robholland:
Closing this. I've confirmed that setting `fullnameOverride` works as expected. Whatever is preventing this from working via Argo must be an Argo issue.

Issue #647, opened 2025-02-17
[Bug] When disabling metrics - chart is not removing of prometheus annotations


itamar-marom:
### What are you really trying to do?

I'm trying to monitor temporal through DataDog autodiscovery option (through annotations)

### Describe the bug

When I'm disabling all metrics annotations there are still some services - the headless services, that still holds the metrics annotations.
Looking at the code I see that there are prometheus labels statically there and indeed cannot be removed through configuration.

![Image](https://github.com/user-attachments/assets/c985cc2e-a573-4dfd-ae6f-db8d62ba03a4)

### Environment/Versions

- OS and processor: EKS amazon linux
- Temporal Version: Chart version 0.54
- Are you using Docker or Kubernetes or building Temporal from source? Kubernetes

robholland:
Does this cause issues in some way?

itamar-marom:
I have a datadog agent that has a the prometheus scraping feature enabled (for other workloads in my cluster). For temporal I'm working with DataDog's temporal integration.
In this current state the datadog agent scrapes tepmoral twice. and in datadog scraping with the prometheus annotations counts as a custom metrics - which costs lost of money

robholland:
I understand, we'll get that fixed.

PR #652, opened 2025-02-25
Turn off service annotations when requested.

robholland:
## What was changed
Added exclusion for the service, as we have for the deployment.

## Why?
Prevents prometheus scraping services when it's not wanted, such as users using Datadog.

## Checklist

1. Closes https://github.com/temporalio/helm-charts/issues/647

2. How was this tested:
Added unit tests.

3. Any docs updates needed?
No.


tomwheeler (approved):
LGTM

PR #654, opened 2025-02-26
Add custom livenessProbe for Temporal services (Feature Request #653)

### LINKED ISSUE #653: [Feature Request] There is no way to configure custom livenessProbe for Temporal services

Sometimes service can loose connectivity to database and it needs to have a possibility to check this event and restart service to reconnect it to DB.
We have faced with this situation during GCP Managed SQL maintenance window when switching to active base can take about 30 seconds.

Liveness Probe like we have below can help us to identify problem and resolve it automatically, but we can't set this custom probe with helm values. It is not supported with current Helm chart.

```livenessProbe:
  exec:
    command:
    - /bin/sh
    - -c
    - tctl --tls_cert_path /etc/temporal/config/certs/cluster.pem --tls_key_path
      /etc/temporal/config/certs/cluster.key --tls_ca_path /etc/temporal/config/certs/ca.cert
      n l
  failureThreshold: 3
  initialDelaySeconds: 150
  periodSeconds: 20
  successThreshold: 1
  timeoutSeconds: 10
```

Please add a possibility to rewrite default livenessProbe configuration via custom Helm values.

Kiseloff:
## What was changed
User can set custom livenessProbe for Frontend/History/Matching services via values.

## Why?
Sometimes service can loose connectivity to database and it needs to have a possibility to check this event and restart service to reconnect it to DB.

## Checklist

1. Closes #653

2. How was this tested:
you can set custom livenessProbe via values.yaml and look at rendered template for deployments.

3. Any docs updates needed?
nothing to update


benwasd:
@tomwheeler , plz merge

thisispr:
@Temporal-Team @temporal-data any update on this? Would be very happy if you could take a look at this.

robholland:
Happy to accept this, please add tests as appropriate though.

PR #659, opened 2025-03-06
Setting busybox repository and tag via values.yaml

FathomlessFunction:
## What was changed
I've updated the helmchart such that the busybox repository and tag are set via values set in values.yaml

## Why?
When working in isolated networks, having this image always set to `busybox:latest` as it is currently will cause the cluster to always try and check for what the latest version is, causing an ImagePullBackoff error. 
Similarly, I've needed to pull this image from a custom registry rather than the public one, too, so having the public repo hard-coded is inconvenient.

It's easy enough to patch yourself, but I thought other users in similar situations would appreciate not needing to do so.

## Checklist

1. Closes N/A

2. How was this tested:
I ran the cluster locally using values.yaml and checked that all pods were healthy.

3. Any docs updates needed?
I don't believe so, but please correct me if I am wrong!

### DIFF
--- a/charts/temporal/templates/server-deployment.yaml
+++ b/charts/temporal/templates/server-deployment.yaml
@@ -45,7 +45,9 @@ spec:
         {{- end }}
         {{- if $.Values.cassandra.enabled }}
         - name: check-cassandra-service
-          image: busybox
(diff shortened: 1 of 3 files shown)


robholland:
We no longer use busybox.

PR #660, opened 2025-03-07
add support for dual visibility

vbatuev:
## What was changed
add helm value for secondary visibility store definition
it used to set up Dual Visibility https://docs.temporal.io/self-hosted-guide/visibility#dual-visibility

## Why?

to set up Dual Visibility in temporal managed by helm 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


robholland:
The support should include the schema management we offer for persistence stores. This is being worked on in: #803.

PR #665, opened 2025-03-12
Add support for annotations on job resource

stephen-harris:
## What was changed
Added support for adding annotations to the schema Job resource. I have leveraged the `temporal.resourceAnnotations` template for consistency. If preferred, we could use the 'direct' approach as used with the annotations in the Job's pod specification, .e.g:

```
  {{- with $.Values.schema.jobAnnotations }}
  annotations:
    {{- toYaml . | nindent 4 }}
  {{- end }}
```

## Why?
We deploy our manifests using ArgoCD and it fails when we make a change to the Job's pod specification (e.g. the secrets being mounted or the environment variables set) as by default ArgoCD uses `apply`. However this behaviour can be reconfigured via [resource annotations](https://argo-cd.readthedocs.io/en/stable/user-guide/sync-options/#replace-resource-instead-of-applying-changes). This PR allows us to set that annotation. 

## Checklist

1. Closes  n/a

2. How was this tested: I ran `helm template` locally 

3. Any docs updates needed? n/a

### DIFF
--- a/charts/temporal/templates/_helpers.tpl
+++ b/charts/temporal/templates/_helpers.tpl
@@ -79,7 +79,7 @@ Create the annotations for all resources
 {{- $scope := index . 1 -}}
 {{- $resourceType := index . 2 -}}
 {{- $component := "server" -}}
-{{- if (or (eq $scope "admintools") (eq $scope "web")) -}}
+{{- if (or (eq $scope "admintools") (eq $scope "web") (eq $scope "schema")) -}}
 {{- $component = $scope -}}
 {{- end -}}
 {{- with $resourceType -}}
--- a/charts/temporal/templates/server-job.yaml
+++ b/charts/temporal/templates/server-job.yaml
@@ -5,6 +5,8 @@ metadata:
   name: {{ include "temporal.componentname" (list $ (printf "schema-%d" .Release.Revision | replace "." "-")) }}
   labels:
     {{- include "temporal.resourceLabels" (list $ "database" "") | nindent 4 }}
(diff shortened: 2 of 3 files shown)


trungdlp:
Any updates @stephen-harris ?

k-ikigai:
Hey! What would be the steps to get it merged?
I can see that there are couple of related issues related to the "schema" job when using argocd and this one has quite a lot of thumbs up :)

Besides current PR I can see one more trying to solve the same: https://github.com/temporalio/helm-charts/pull/743
but according to discussion in this issue: https://github.com/temporalio/helm-charts/pull/555, helm hooks were already present at some point in time.

And one more related feature request: https://github.com/temporalio/helm-charts/issues/735

But.. on the topic again. This approach seems to be the least intrusive and safest one. If there is something more to be done I am glad to further improve this PR, but would need to know what exactly. @robholland could you guide me?

jmooo:
@robholland is there something additional we'd need to do to get this small PR reviewed? 🙏

robholland:
Hey, sorry for the long delay here. Please can you update this after the changes we made for v1. It should be a very easy merge in your case as it's only the schema block in values I think.

Once that's done I will get this merged.

robholland:
Oh, also, please add a test for it.

k-ikigai:
Hey @robholland thank you for response. I created my own PR to be able to resolve everything you mentioned. Please check this one: https://github.com/temporalio/helm-charts/pull/815

robholland:
Closing in favour of #815

PR #672, opened 2025-03-16
[server-job] update TEMPORAL_ADDRESS

luismacosta:
[server-job] update TEMPORAL_ADDRESS

## What was changed
server-job.yaml

For InitContainer _create-{{ $namespace.name }}-namespace_  allow env var TEMPORAL_ADDRESS as ingress endpoint

### DIFF
--- a/charts/temporal/templates/server-job.yaml
+++ b/charts/temporal/templates/server-job.yaml
@@ -147,6 +147,11 @@ spec:
             - name: TEMPORAL_ADDRESS
               {{- if and (hasKey $.Values.server "internalFrontend") $.Values.server.internalFrontend.enabled }}
               value: {{ include "temporal.fullname" $ }}-internal-frontend.{{ $.Release.Namespace }}.svc:{{ $.Values.server.internalFrontend.service.port }}
+              {{- else if $.Values.server.frontend.ingress.enabled }}
+              {{- range $.Values.server.frontend.ingress.hosts }}
+              {{- $url := splitList "/" . }}
+              value: "{{ first $url }}:{{ $.Values.server.frontend.service.port }}"
+              {{- end }}
               {{- else }}
               value: "{{ include "temporal.fullname" $ }}-frontend.{{ $.Release.Namespace }}.svc:{{ $.Values.server.frontend.service.port }}"
               {{- end }}


ashwin153:
@robholland I think I'm also running it this bug. It prevents the namespaces from being created.

robholland (changes requested):
This needs to pick the first host, not loop over each of them. As-is you would print `value: ...` multiple times if there are multiple hosts.

luismacosta:
> This needs to pick the first host, not loop over each of them. As-is you would print `value: ...` multiple times if there are multiple hosts.

I'm using {{ first $url }}

charts/temporal/templates/server-job.yaml
```diff
@@ -147,6 +147,11 @@ spec:
             - name: TEMPORAL_ADDRESS
               {{- if and (hasKey $.Values.server "internalFrontend") $.Values.server.internalFrontend.enabled }}
               value: {{ include "temporal.fullname" $ }}-internal-frontend.{{ $.Release.Namespace }}.svc:{{ $.Values.server.internalFrontend.service.port }}
+              {{- else if $.Values.server.frontend.ingress.enabled }}
+              {{- range $.Values.server.frontend.ingress.hosts }}
```
robholland, line 151: You are using `range` here. That's a loop over each frontend ingress host, which will result in `value: ..` bring printed multiple times.

robholland, line 151: You are using `range` here. That's a loop over each frontend ingress host, which will result in value: .. bring printed multiple times. The `first $url` code picks out the domain name by splitting the full URL by `/` and taking the first item in that list, it's not related to picking the first ingress host.

robholland (changes requested):

charts/temporal/templates/server-job.yaml
luismacosta: ok got it.
so, I should remove the range and define only for instance Values.server.frontend.ingress.host

robholland: I suggest using the first value in `$.Values.server.frontend.ingress.hosts`.

robholland: I'm curious why the server job can't talk to the frontend service directly rather than through the ingress though?

charts/temporal/templates/server-job.yaml
```diff
@@ -147,6 +147,8 @@ spec:
             - name: TEMPORAL_ADDRESS
               {{- if and (hasKey $.Values.server "internalFrontend") $.Values.server.internalFrontend.enabled }}
               value: {{ include "temporal.fullname" $ }}-internal-frontend.{{ $.Release.Namespace }}.svc:{{ $.Values.server.internalFrontend.service.port }}
+              {{- else if $.Values.server.frontend.ingress.enabled }}
+              value: "{{ $.Values.server.frontend.ingress.temporaladdress }}"
```
luismacosta, line 151:
```suggestion
              value:  "{{ index $.Values.server.frontend.ingress.hosts 0 }}"
```

charts/temporal/values.yaml
```diff
@@ -214,6 +214,7 @@ server:
       httpPort: 7243
     ingress:
       enabled: false
+      #temporaladdress:
```
luismacosta, line 217:
```suggestion
```

luismacosta:
Hello @robholland can you please review again? Thanks.

robholland (approved):

PR #676, opened 2025-03-21
Closes #675: Enable persistence to the default Cassandra and ES

### LINKED ISSUE #675: [Feature Request] FILL_TITLE_HERE

I'm using the default helm chart to test temporal both locally and in our test k8s cluster. However, the current default chart are using transient storage for ES and Cassandra.
This results in the cluster being down each time the ES/Cassandra node is killed. The reason is that schema migration is only run once at the time of initialization, so new ES/Cassandra node don't have the needed schemas.
There is no way to recover from this except starting a new deployment.

Enable persistence to Cassandra and ES config.

hungcq:
## What was changed
Enable persistence to the default Cassandra and ES

## Why?
The current default chart are using transient storage for ES and Cassandra.
This results in the cluster being down each time the ES/Cassandra instances are killed. The reason is that schema migration is only run once at the time of initialization, so new ES/Cassandra instances don't have the needed schemas.
There is no way to recover from this except starting a new deployment.

## Checklist

1. Closes #675 

2. How was this tested:
I deployed a Temporal cluster with the new configs, and then killed the Cassandra/ES instance. After that, I checked whether the cluster still worked properly.

3. Any docs updates needed?
N/A

### DIFF
--- a/charts/temporal/values.yaml
+++ b/charts/temporal/values.yaml
@@ -451,7 +451,7 @@ elasticsearch:
   enabled: true
   replicas: 3
   persistence:
-    enabled: false
+    enabled: true
   imageTag: 7.17.3
   host: elasticsearch-master-headless
   scheme: http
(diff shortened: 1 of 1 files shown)


robholland:
From v1 rc1 onward we don't install any persistence backends on the user's behalf.

Issue #664, opened 2025-03-12
[Bug] admintools exposes non-existing SSH on Pod and Service


Drahflow:
### What are you really trying to do?

Configuring a minimal set of required ports in our temporal k8s setup for some security product.

### Describe the bug

[The admintool pod exposes an SSH port (called `http`)](https://github.com/temporalio/helm-charts/blob/195bb8b74b66df54442e76fe66649a68468a7db6/charts/temporal/templates/admintools-deployment.yaml#L34-L36) which is also advertised [as a service](https://github.com/temporalio/helm-charts/blob/195bb8b74b66df54442e76fe66649a68468a7db6/charts/temporal/templates/admintools-service.yaml).

However, there is nothing actually listening on 22 in the container.

### Minimal Reproduction

`kubectl -n temporal port-forward temporal-admintools-<your pod here> 2222:22`
`netcat 127.0.0.1 2222` and observe the connection errors in `kubectl`

### Environment/Versions

This is on current temporal, deployed via helm-chart in kubernetes.

### Additional context

The port-22 exposing code seems to be here since the initial commit. It's likely some SSH server existed in earlier times.

PR #677, opened 2025-03-21
Remove admintools-service.

robholland:
It looks like it was intended to allow SSH into the admintools pods, but they don't run sshd or have any ssh keys on them.


luismacosta (approved):

robholland:
Fixes #664

Drahflow:
The pod still declares a port `22`. Was it intentional to keep that part?

tomwheeler (approved):
I tested this and found that the port is no longer exposed (specifically by using the ssh/netcat approach mentioned in #664)

antoncohen:
The Service isn't only for SSH, it allowed for easy use of `kubectl exec`. The usage of the admintools Service is [documented in the README](https://github.com/temporalio/helm-charts?tab=readme-ov-file#running-temporal-cli-from-the-admin-tools-container):

```
kubectl exec -it services/temporaltest-admintools /bin/bash
```

I believe that could be changed to:

```
kubectl exec -it deployment/temporaltest-admintools -- /bin/bash
```

PR #678, opened 2025-03-21
Allow adjusting log values

johnkrzywanek:
## What was changed

Allow setting log values like output file from the helm chart values

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


robholland:
Please add tests as appropriate and don't add config lines when nothing has been set (can use `with` to do that cleanly).

johnkrzywanek:
Thanks @robholland! Just adjusted the code

robholland (approved):

Issue #680, opened 2025-03-25
[Feature Request] Add ability to make temporal-web service headless


abarganier:
### Is your feature request related to a problem? Please describe.

We have a need to make the `temporal-web` service headless so that it can be targeted by our external load balancers.

### Describe the solution you'd like

Add the ability to set `spec.clusterIP: None` in the [web-service.yaml](https://github.com/temporalio/helm-charts/blob/main/charts/temporal/templates/web-service.yaml). By default, this can be disabled, but it will give users the ability to turn the service into a headless one if need be.

If the service is headless, there's probably no reason to keep the standard ClusterIP version around, so this can be a "one or the other" type setup.

abarganier:
I'll file a PR with these changes.

PR #681, opened 2025-03-25
Add Values.web.service.clusterIP to provide headless temporal-web service toggle

abarganier:
## What was changed
Addresses: https://github.com/temporalio/helm-charts/issues/680

This PR adds support for `Values.web.service.clusterIP`, which can be set to `None` to turn the `temporal-web` service into a headless service. 

## Why?
We need to be able to target the service with an external load balancer, which requires the service to be headless. This small change gives users the capability to convert the service into a headless one.

## Checklist
1. Closes https://github.com/temporalio/helm-charts/issues/680

2. How was this tested:
[x] Added `charts/temporal/tests/web_service_test.yaml` with new unit tests to test the new value.

3. Any docs updates needed?
- Inline comment docs were added above the new value to indicate that it's supported and explain its usage.

### DIFF
--- a/charts/temporal/templates/web-service.yaml
+++ b/charts/temporal/templates/web-service.yaml
@@ -13,6 +13,9 @@ spec:
   loadBalancerIP: {{.}}
     {{- end }}
   type: {{ .Values.web.service.type }}
+  {{- with .Values.web.service.clusterIP }}
+  clusterIP: {{.}}
+  {{- end }}
   ports:
     - port: {{ .Values.web.service.port }}
       targetPort: http
--- a/charts/temporal/tests/web_service_test.yaml
+++ b/charts/temporal/tests/web_service_test.yaml
@@ -1,0 +1,31 @@
+suite: test server service
+templates:
+  - web-service.yaml
+tests:
+  - it: includes the provided clusterIP to make the service headless
+    template: templates/web-service.yaml
+    documentSelector:
+      path: kind
+      value: Service
+    set:
+      web:
+        service:
+          clusterIP: None
+    asserts:
+      - equal:
+          path: spec.type
+          value: ClusterIP
+      - equal:
+          path: spec.clusterIP
+          value: None
+  - it: does not include a spec.clusterIP by default, when none is given
+    template: templates/web-service.yaml
+    documentSelector:
+      path: kind
+      value: Service
+    asserts:
+      - equal:
+          path: spec.type
+          value: ClusterIP
+      - notExists:
+          path: spec.clusterIP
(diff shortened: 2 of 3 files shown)


charts/temporal/templates/web-service.yaml
```diff
@@ -13,6 +13,9 @@ spec:
   loadBalancerIP: {{.}}
     {{- end }}
   type: {{ .Values.web.service.type }}
+  {{- if .Values.web.service.clusterIP }}
```
robholland, line 16: Please use `with` rather than `if`.

charts/temporal/templates/web-service.yaml
```diff
@@ -13,6 +13,9 @@ spec:
...
+  clusterIP: {{ .Values.web.service.clusterIP }}
+  {{- end}}
```
robholland, line 18:
```suggestion
  {{- end }}
```

robholland (changes requested):

charts/temporal/tests/web_service_test.yaml
```diff
@@ -0,0 +1,37 @@
...
+          value: ClusterIP
+      - equal:
+          path: spec.clusterIP
+          value: None
+  - it: does not include a spec.clusterIP by default, when none is given
+    template: templates/web-service.yaml
+    documentSelector:
+      path: kind
+      value: Service
+      matchMany: true
```
robholland, line 29: No need for matchMany.

charts/temporal/tests/web_service_test.yaml
```diff
@@ -0,0 +1,37 @@
...
+  - web-service.yaml
+tests:
+  - it: includes the provided clusterIP to make the service headless
+    template: templates/web-service.yaml
+    documentSelector:
+      path: kind
+      value: Service
+      matchMany: true
+    values:
+      - ../values.yaml
```
robholland, line 12: You don't need to set `values`, it will use the default file.

charts/temporal/templates/web-service.yaml
abarganier: Done.

charts/temporal/templates/web-service.yaml
abarganier: Done.

charts/temporal/tests/web_service_test.yaml
abarganier: Done.

charts/temporal/tests/web_service_test.yaml
abarganier: Done.

abarganier:
Thanks for the review! Addressed your feedback & ready for another look.

PR #690, opened 2025-04-10
Add chart cassandra

luismacosta:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


charts/temporal/templates/_helpers.tpl
```diff
@@ -432,6 +450,7 @@ Renders a value that contains template.
 Usage:
 {{ include "common.tplvalues.render" ( dict "value" .Values.path.to.the.Value "context" $) }}
 */}}
+
```
luismacosta, line 453:
```suggestion
```

luismacosta:
> Nice work @luismacosta!
> 
> I have some feedback for you: The commands that you provided needed to be tweaked, in the future to ease PR testing please ensure that your commands are repeatable without change.
> 
> ```shell
> $ cd charts/temporal
> $ helm dependency build
> $ helm template . --name-template temporal -n temporal-tests -f ci/cassandra.yaml  --debug  > final.yaml
> $ kubectl create ns temporal-tests
> $ kubectl apply -f final.yaml -n temporal-tests
> ```
> 
> Additionally I have a few questions:
> 
> 1. Why are you setting up MySQL as the visibility store instead of using Cassandra for both? I don't see anything in the feature request that would suggest this change is needed / wanted by Temporal.
> 2. Why are you using `jeremias` as the root password? Look to use more standard naming conventions such as "temoral" or "temporalRoot"
> 3. I think it's fine that you updated the ci/cassandra but I don't see you make mention of the `values/values.cassandra.yaml`. Was this tested with the Bitnami chart?
> 
> Also as this is a helm-chart repo, please use helm to install and manage the deployment.

Hello @PhillypHenning 

Thanks for the feedback.

1. Mysql as visibility store was defined just to test the helm chart
2. Password has been updated to temporalRoot
3. values.cassandra.yaml updated

luiscosta18:
Hello @PhillypHenning

I've updated the PR with your suggestions.
Can you please review it and merge if everything looks good to you? Thanks.

Cc @robholland @tomwheeler

charts/temporal/Chart.yaml
```diff
@@ -28,6 +28,10 @@ dependencies:
     repository: https://grafana.github.io/helm-charts
     version: 8.0.2
     condition: grafana.enabled
+  - name: mysql
```
PhillypHenning, line 31: I think this is out-of-scope for the feature request. The only change that should be made here is updating the cassandra dependency to Bitnami from incubator. 

I see a lot of additions here for MySQL that I don't think are required. 

I would recommend scaling back the changes to only that which satisfies the feature request in [Milestone 1](github.com/temporalio/helm-charts/issues/468)

charts/temporal/Chart.lock
```diff
@@ -11,5 +11,8 @@ dependencies:
 - name: grafana
   repository: https://grafana.github.io/helm-charts
   version: 8.0.2
-digest: sha256:f74565632d43941cad32e713f73481b3d8065d9c462473e80d86fd1f7c5049af
-generated: "2024-10-07T14:43:06.775398+01:00"
+- name: mysql
+  repository: oci://registry-1.docker.io/bitnamicharts
+  version: 12.3.3
+digest: sha256:5cffc6a0f802a924ac7bda6597a51831fd082fe1208813faeaf4c12873283aa9
+generated: "2025-04-10T03:14:27.145429+01:00"
```
luiscosta18, line 18:
```suggestion
```

luiscosta18 (changes requested):

sergey-vin:
This is probably a bad idea. bitnami is now longer supporting OCI images:
https://medium.com/@talkimhi/bitnamis-august-28th-bombshell-the-end-of-free-container-images-as-we-know-them-74fe5cdfb882

PR #691, opened 2025-04-10
Make schema job ttl configurable

peter-c-larsson:
## What was changed
Making job ttl configurable with default values to be backwards compatible

## Why?
We ran into this issue when deploying the helm chart with argocd:
https://github.com/argoproj/argo-cd/issues/8586

## Checklist

1. Closes 

2. How was this tested:
Test it just making a deploy of the helm chart make sure it have been intact. Will test the changes myself as well.

3. Any docs updates needed?

### DIFF
--- a/charts/temporal/templates/server-job.yaml
+++ b/charts/temporal/templates/server-job.yaml
@@ -7,7 +7,9 @@ metadata:
     {{- include "temporal.resourceLabels" (list $ "database" "") | nindent 4 }}
 spec:
   backoffLimit: {{ $.Values.schema.setup.backoffLimit }}
-  ttlSecondsAfterFinished: 86400
+  {{- if $.Values.schema.jobttl.enabled }}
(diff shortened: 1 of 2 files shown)


tKe:
An alternative would be to allow for custom annotations on the Job.
This would allow for
```yaml
argocd.argoproj.io/hook-delete-policy: HookSucceeded
argocd.argoproj.io/sync-options: Force=true,Replace=true
```
to properly clean up in a way that Argo won't complain about sync, and allow for changes in the job should it have failed.

With disabling the ttl, all you gain is the sync no longer complaining. However a lot of fields on a Job are immutable (most notably the entire `.spec.template`) and so future syncs are very likely to fail as the Job still exists and won't permit update via the standard argo apply approach (hence setting `Replace=true` to cover the scenario of a sync with changes to the template after a previous Job failure and no subsequent removal due to `HookSucceded`).

peter-c-larsson:
Thanks for the update and closing this MR in favor of the other one. Lets hope that MR can be merged and released soon.

PR #692, opened 2025-04-14
[Feature Request] add keda worker autoscaler based on targetQueueSize

luismacosta:
[Feature Request] add keda worker autoscaler based on targetQueueSize

@robholland @tomwheeler , Can you please review? Thanks

## What was changed
Add keda-scaledobject.yaml to templates

## Why?
keda v.2.17.0 contains a Temporal Worker scaler.
This solution empowers developers to dynamically scale the number of Temporal workers based on the size of the Task Queue backlog.

Info:
https://temporal.io/blog/announcing-keda-based-auto-scaling-for-temporal-workers 
https://keda.sh/docs/2.17/scalers/temporal/ 
https://github.com/kedacore/keda/releases/tag/v2.17.0 

## How was this tested:
➜  helm template .  --name-template temporal -n temporal-tests -f ci/keda.yaml   > final.yaml --debug

### DIFF
--- a/charts/temporal/values.yaml
+++ b/charts/temporal/values.yaml
@@ -348,6 +348,55 @@ server:
     containerSecurityContext: {}
     topologySpreadConstraints: []
     podDisruptionBudget: {}
+    keda:
+      enabled: true
+      pollingInterval: 30
+      cooldownPeriod: 300
+      initialCooldownPeriod: 0
+      minReplicaCount: 0
+      maxReplicaCount: 5
+      fallback: {}
+      # failureThreshold: 3
+      # replicas: 6
+      advanced: {}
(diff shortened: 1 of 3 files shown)


charts/temporal/values.yaml
```diff
@@ -348,6 +348,55 @@ server:
     containerSecurityContext: {}
     topologySpreadConstraints: []
     podDisruptionBudget: {}
+    keda:
+      enabled: true
```
luismacosta, line 352:
```suggestion
      enabled: false
```

robholland:
Thank you for the contribution, but the Helm chart is not responsible for application workers. Configuring a KEDA scaler is out of scope for the Helm chart and should be handled by a user's worker deployment mechanism.

Issue #695, opened 2025-04-16
[Feature Request] Decouple server job manifests from `Values.server.enabled`


abarganier:
### Is your feature request related to a problem? Please describe.

Running the schema migration jobs is something that we'd like to trigger independently from normal server deployments. Schema migrations are only needed in a couple cases:
1. When setting up a new cluster.
2. When performing a version upgrade.

Therefore, we want fine control over when these jobs get deployed, **separate** from the rest of the server Deployments. 

However, currently the server job manifests do not get generated if `Values.server.enabled: false`, meaning we're unable to isolate the server job from the core service manifests themselves.

### Describe the solution you'd like

Decouple the server job from the rest of the server manifests, so that we can properly isolate the job into a separate deployment. 

### Additional context

Note that the server job depends on the secrets created containing DB credentials. Therefore, we need to make sure that these secrets are still created when `Values.server.enabled: false`, but any one of the `schema` values are enabled.

PR #696, opened 2025-04-16
Decouple server job manifests from Values.server.enabled

abarganier:
## What was changed
Addresses: https://github.com/temporalio/helm-charts/issues/695

This PR allows the server job manifests to be generated even when `Values.server.enabled: false`. Previously, it was not possible to isolate the server job manifests from the rest of the server Deployments (e.g. History, Matching, Frontend, etc).

Additionally, it updates the logic in the server-secret template to ensure that the secrets manifests are still generated when `Values.server.enabled: false`, but one of the schema jobs is enabled. This ensures that the server job still has access to the DB credential secrets required to operate. 

## Why?
Running the schema migration jobs is something that we'd like to trigger independently from normal server deployments. Schema migrations are only needed in a couple cases:
1. When setting up a new cluster.
2. When performing a version upgrade.

Therefore, we want fine control over when these jobs get deployed, separate from the rest of the server Deployments.

## Checklist
1. Closes https://github.com/temporalio/helm-charts/issues/695

2. How was this tested:
[x] Created additional unit tests

### DIFF
--- a/charts/temporal/templates/server-job.yaml
+++ b/charts/temporal/templates/server-job.yaml
@@ -1,4 +1,4 @@
-{{- if $.Values.server.enabled }}{{- if or $.Values.schema.createDatabase.enabled $.Values.schema.setup.enabled $.Values.schema.update.enabled }}
+{{- if or $.Values.schema.createDatabase.enabled $.Values.schema.setup.enabled $.Values.schema.update.enabled }}
 apiVersion: batch/v1
 kind: Job
 metadata:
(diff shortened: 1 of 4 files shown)


tomwheeler (approved):
I tested it to verify that it worked as described (all tests) pass and also validated that I could deploy a cluster using this code (there were no regressions or unexpected behaviors).

tomwheeler:
Thanks for your contribution (and especially for including the unit tests)!

I plan to release a new version of helm-charts on Monday morning (US time) that will include this.

abarganier:
Thanks for the review, very much appreciate your help!

Issue #698, opened 2025-04-23
[Bug] Schema Job failed to create namespace


msolimans:
### What are you really trying to do?

I tried to install temporal with helm chart values and I turned on namespace's creation to `true` but schema job in init container `create-x-namespace` 

### Describe the bug

running 
here is the code snippet from values.yml I used

```
server:
  config:
    namespaces:
      create: true
      namespace: 
        - name: temporalx 
          retention: 3d
```

schema pod crashed in creating namespace - here are the logs

```
kubectl logs -n temporal releease-name-temporal-scehma.... -c create-temporalx-namespace

Error: provide either namespace-id flag or namespace as an argument. Global flag 'namespace' is not supported by namespace commands

```

### Minimal Reproduction

use `helm template` to generate manifests and apply them.
- helm repo add temporal ...
- helm repo update 
- helm template .. > temporal.yml 

### Environment/Versions

- OS and processor: [e.g. M1 Mac, x86 Windows, Linux]
- Temporal Version: [e.g. 1.14.0?] and/or SDK version
- Are you using Docker or Kubernetes or building Temporal from source?

robholland:
I can't replicate with the current chart. Please re-open if you see this again.

PR #699, opened 2025-04-23
fix: namespace creation with temporal operator

msolimans:
removed -n argument to fix issue that is reported here https://github.com/temporalio/helm-charts/issues/698

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


robholland:
Which admintools image is being used in your deployment? `-n` is the correct command for the current CLI release: https://github.com/temporalio/cli/blob/main/temporalcli/commands.operator_namespace.go#L22

robholland:
Please re-open if you confirm an issue with current versions.

PR #708, opened 2025-05-26
Allow usage of HPAs

purkhusid:
## What?
Remove the default replica values of `1` since that clashes with the usage of HPAs. Instead default to not setting `replicas`, which by default is `1`.


robholland:
Makes sense. Please add a test.

robholland (changes requested):

purkhusid:
@robholland This should be ready

robholland:
There are still comments not attended to.

purkhusid:
I guess I got sidetracked but I've now addressed all the comments @robholland

robholland:
This is failing the unit tests, please ensure those pass.

purkhusid:
@robholland Tests are fixed

robholland (approved):

PR #712, opened 2025-06-05
Add readiness probes

### LINKED ISSUE #710: [Bug] Missing readiness probes on web server

Send requests to temporal web server only when it's ready to accept requests.

### Describe the bug

As there are missing readiness we cannot properly load balance requests on startup or pod restart as we will forward request to pods not really ready to serve requests.

mtparet:
## What was changed
Adding readiness probes to web and server deployment.

## Why?
Currently we send requests to these pods even when these are not fully started. These basic readiness should help wait before sending requests that temporal is up.

## Checklist

1. Closes #710

### DIFF
--- a/charts/temporal/templates/server-deployment.yaml
+++ b/charts/temporal/templates/server-deployment.yaml
@@ -135,6 +135,10 @@ spec:
              initialDelaySeconds: 150
              tcpSocket:
                port: rpc
+          readinessProbe:
+              initialDelaySeconds: 10
+              tcpSocket:
+                port: rpc
           {{- end }}
           volumeMounts:
             - name: config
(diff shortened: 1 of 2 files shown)


VLZZZ:
Hey team,
Is this going to be reviewed/merged anytime soon?

We experiencing drops anytime we upgrade Temporal cluster with multiple replicas.
While we're following `RollingUpdate` strategy and rolling out performed `1 by 1`, I've noticed that pod immediately reaches `Ready` status while logs still states that:
`"operation":"PollActivityTaskQueue","wf-namespace":"<redacted>","error":"Frontend is not healthy yet"`

And only several seconds (or faster) later
`Frontend is now healthy`

Under the heavy load we encountering too many rejects which affecting performance when using Temporal

Thank you!

robholland:
Happy to add this, but they should be configurable. Please also add tests.

quangngotan95:
@robholland as we need this, I opened another PR here https://github.com/temporalio/helm-charts/pull/825, please help review it 🙇‍♂️

robholland:
Implemented in #825

PR #715, opened 2025-06-11
Adds minReadySeconds supports for servers

quangngotan95:
## What was changed
This PR adds supports for `minReadySeconds` on server components
## Why?
When we perform any update to Temporal, it rotates Temporal pods and shard-rebalancing kicks in.
The current Chart makes rotating pods too quick and the system became unstable. We get many 5xx and latency spikes during that time.
By adding this parameter and slow down pods rotation a bit, we could slow down shard-rebalancing and help stablize the system.
## Checklist

1. Close
N/A

2. How was this tested: We've been running this in production for almost a year with our fork.

4. Any docs updates needed?
N/A

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @temporalio/selfhosted
+* @temporalio/selfhosting


mtparet:
There is not readiness probes at all so we currently we try serving not ready pods. Readiness probes should be enough.

PR #716, opened 2025-06-12
feat: add possibility to bring your own server configmap

space192:
## What was changed
An option to disable the default config map was added.

## Why?
The idea is to allow the Temporal Helm chart to be used as a subchart, and by doing so we can bring our own config map with custom templating for the database/bucket.
The dynamic config map is great but it doesn't allow for dynamic templating inside of it.

2. How was this tested:
This doesn't change anything for the regular user; people toggling this should know what they are doing.


space192:
@lucasmrdt (as we are working together)

PR #721, opened 2025-06-30
fix: Add "v" prefix on ui image tag

### LINKED ISSUE #720: [Bug] Web component docker image url returns 404

Install 0.64.0 version of the Temporal Helm chart.

### Describe the bug

Starting from version 2.37.2, ui team is using **v** as version prefix, and therefore the images are someting like this `temporalio/ui:v2.37.3` (check here https://hub.docker.com/r/temporalio/ui/tags).

This is causing the Web component not to start because in the values.yaml file `.web.image.tag` is still using the previous format 2.37.2.

### Minimal Reproduction

Install the chart with all default values (or update from a previous verision).

karatekaneen:
## What was changed

Changed image tag for UI deployment

## Checklist

1. Closes #720 

2. How was this tested:
Did a override with the new default value in our deployment

3. Any docs updates needed?
No

### DIFF
--- a/charts/temporal/values.yaml
+++ b/charts/temporal/values.yaml
@@ -381,7 +381,7 @@ web:
   replicaCount: 1
   image:
     repository: temporalio/ui
-    tag: 2.38.3
+    tag: v2.38.3
     pullPolicy: IfNotPresent
   service:
     # set type to NodePort if access to web needs access from outside the cluster


robholland:
This was due to a regression in our docker image tagging. This is fixed now, and missing tags have been pushed. Thanks for reporting.

PR #722, opened 2025-07-02
Templates now have metadata.namespace defined

Decard6:
## What was changed
Templates now have `metadata.namespace` defined to `$.Release.Namespace`

## Why?
Leaving namespace undefined works while using `helm install -n my-namespace`. When using `helm template -n my-namespace` however the namespace will not appear in the yamls.

1. Closes
I have not created an issue for this.

2. How was this tested:
By templating the chart.

3. Any docs updates needed?
With no new values - probably not.


Decard6:
I tried porting Bitnami's `common.names.namespace` but it's not so simple due to how your templates are written. The `namespaceOverride` feature is not a priority to me so I scrapped that and left `$.Release.Namespace` only in the templates.

PR #724, opened 2025-07-10
fix: web ui uses internalFrontend when available

### LINKED ISSUE #706: [Feature Request] Temporal UI use internal-frontend

Hey, could we enable the web deployment to use the internal-frontend address?

golgoth31:
## What was changed
This pull request allow the web ui to use the internal frontend when it's available.

## Why?
With the current configuration the web ui is only linked to the common frontend, when anabling authorization on this frontend, the UI is enable to work properly. Using the internal frontend allows using a different OIDC provider for jwt access then the jwt authorization in the frontend.

## Checklist

1. Closes #706 

2. How was this tested:
deployed and working in our system.

3. Any docs updates needed?

### DIFF
--- a/charts/temporal/templates/web-deployment.yaml
+++ b/charts/temporal/templates/web-deployment.yaml
@@ -36,7 +36,11 @@ spec:
           imagePullPolicy: {{ .Values.web.image.pullPolicy }}
           env:
             - name: TEMPORAL_ADDRESS
+              {{- if and (hasKey .Values.server "internalFrontend") .Values.server.internalFrontend.enabled .Values.web.useInternalFrontend }}
+              value: "{{ include "temporal.fullname" $ }}-internal-frontend.{{ .Release.Namespace }}.svc:{{ .Values.server.internalFrontend.service.port }}"
+              {{- else }}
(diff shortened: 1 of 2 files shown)


charts/temporal/templates/web-deployment.yaml
```diff
@@ -36,7 +36,11 @@ spec:
           imagePullPolicy: {{ .Values.web.image.pullPolicy }}
           env:
             - name: TEMPORAL_ADDRESS
+              {{- if and (hasKey .Values.server "internalFrontend") .Values.server.internalFrontend.enabled .Values.web.useInternalFrontend }}
```
robholland, line 39:
```suggestion
              {{- if and (index .Values.server "internal-frontend").enabled .Values.web.useInternalFrontend }}
```

charts/temporal/templates/web-deployment.yaml
```diff
@@ -36,7 +36,11 @@ spec:
...
+              value: "{{ include "temporal.fullname" $ }}-internal-frontend.{{ .Release.Namespace }}.svc:{{ .Values.server.internalFrontend.service.port }}"
```
robholland, line 40:
```suggestion
              value: {{ include "temporal.fullname" $ }}-internal-frontend:{{ (index $.Values.server "internal-frontend").service.port }}
```

robholland (changes requested):

Issue #731, opened 2025-08-01
[Bug] Elasticsearch jobs/ checks fail for clusters with no users auth


quangngotan95:
### What are you really trying to do?
I'm trying to install Temporal with the Charts.
It's failing for ES cluster with no auth (no username/ password required)

### Describe the bug
For ES cluster with no auth (no username/ password required), it'll fail since we always have `--user "$ES_USER:$ES_PWD" in our `curl`

https://github.com/temporalio/helm-charts/blob/22413b6c9e1e38f7b32e1278cf3de173397bc166/charts/temporal/templates/server-job.yaml#L44

Reverting to checks before [this change](https://github.com/temporalio/helm-charts/pull/530) would resolve it.

### Minimal Reproduction
Install the Charts with ES store but do not supply `username` and `password`

### Environment/Versions
Charts Version: 0.64.0

- OS and processor: [e.g. M1 Mac, x86 Windows, Linux]
- Temporal Version: [e.g. 1.14.0?] and/or SDK version
- Are you using Docker or Kubernetes or building Temporal from source?

PR #732, opened 2025-08-02
fix: ES cluster access with no user auth

quangngotan95:
## What was changed

This PR modifies `curl` requests to ES to support clusters with no user auth.
Fix https://github.com/temporalio/helm-charts/issues/731

## Why?

ES clusters with no user auth are not working

## Checklist

1. Closes 
https://github.com/temporalio/helm-charts/issues/731

2. How was this tested:

Helm template and install to our k8s cluster.

3. Any docs updates needed?


quangngotan95:
@PhillypHenning thank you, I wasn't using the dependent ES, it's been fixed now

robholland:
This is fixed from v1 rc1 onward.

PR #746, opened 2025-09-16
feat: add values.schemas.json

mathewjhan:
## What was changed

Added a values-schemas-json generated by `helm plugin install https://github.com/losisin/helm-values-schema-json.git && helm schema`. 

## Why?

This is useful for some applications such as cdk8s which generates the constructs based on this schema. It may be worth adding a Github action or precommit to autogenerate this. 

[see: README](https://github.com/losisin/helm-values-schema-json)

## Checklist

1. Closes 

2. How was this tested:

N/A

3. Any docs updates needed?

N/A

### DIFF
--- a/charts/temporal/values.schema.json
+++ b/charts/temporal/values.schema.json
@@ -1,0 +1,1367 @@
+{
+    "$schema": "https://json-schema.org/draft/2020-12/schema",
+    "type": "object",
+    "properties": {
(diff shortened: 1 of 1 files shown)


robholland:
Unfortunately there isn't a lot of value (pun not intended) in merging this, it's already out of date. If you want to file an issue we can look at automating this though.

PR #754, opened 2025-10-02
Use temporal cli instead of tctl

Qreepex:
## What was changed

Use the Temporal CLI instead of tctl in the README.md for the example commands.

## Why?

tctl is deprecated and EOS

## Checklist

1. Closes 

none

3. How was this tested:

Executed the commands in temporal and they worked

4. Any docs updates needed?

its kinda a docs update


robholland:
Fixed in #793

Issue #735, opened 2025-08-08
[Feature Request] Allow Schema Job ttlSecondsAfterFinished to be configurable


PurseChicken:
### Is your feature request related to a problem? Please describe.

Currently the ttlSecondsAfterFinished spec is hard coded to 86400. It would be ideal if not only you could change this value, but you could remove it altogether.

This is especially important if you use Argo for deployment. Eventually the after-finished controller kicks in and then the job disappears. Argo then reports that there is something missing in that particular application until its synced again.

### Describe the solution you'd like

This should be easily achievable by adding some key value pairs in the values file and key off of those in the job template.

values.yaml:
```
schema:
  garbageCollection:
    # Enable or disable the job garbage collection
    enabled: true
    # Time before the job is removed in seconds Default 86400
    ttlSeconds: 86400
```

templates/server-job.yaml
existing:
```
spec:
  backoffLimit: {{ $.Values.schema.setup.backoffLimit }}
  ttlSecondsAfterFinished: 86400
```

updated:
```
spec:
  backoffLimit: {{ $.Values.schema.setup.backoffLimit }}
  {{- if and $.Values.schema.garbageCollection $.Values.schema.garbageCollection.enabled }}
  ttlSecondsAfterFinished: {{ $.Values.schema.garbageCollection.ttlSeconds | default 86400 }}
  {{- end }}
```

PurseChicken:
Bump this. Do you want me to create a PR?

PurseChicken:
Bumping again. Could really use this.

danifuuu:
@PurseChicken talking about managing temporal using ArgoCD and this job, aren't you running into issues while updating the Helm chart? I am facing inmutable fields regarding the job which is quite normal but I was wondering how do you manage updates

PurseChicken:
@danifuuu Yes. When updating the chart, you just need to delete the job resource before syncing and then everything is fine.

danifuuu:
> [@danifuuu](https://github.com/danifuuu) Yes. When updating the chart, you just need to delete the job resource before syncing and then everything is fine.

but manually? seems a bit meh

PurseChicken:
> > [@danifuuu](https://github.com/danifuuu) Yes. When updating the chart, you just need to delete the job resource before syncing and then everything is fine.
> 
> but manually? seems a bit meh

I agree. You can also use sync options which would help in this situation as well.. but its more of a workaround:

```
[...]

PR #761, opened 2025-10-17
feat: Allow the ability to configure garbage collection on schema-job.

PurseChicken:
## What was changed

This PR gives the user the ability to configure schema-job garbage collection. Specifically the ability to enable or disable the garbage collection altogether as well as specify the desired time before garbage collection runs against the job if enabled.

See [Here](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/) for more information about kubernetes garbage collection on job resources.

## Why?
Currently garbage collection (.spec.ttlSecondsAfterFinished) is hard coded in the server-job manifest. This means that the job will always delete itself after the specified value (currently 86400). In some cases users may want to either configure the number of seconds before deleting the job manifest, or they may want to disable this garbage collection altogether. This PR takes care of both of those scenarios.

## Checklist

2. How was this tested:
The chart was validated using helm template --debug as well as packaging it (helm package) and using it in a deployment.

3. Any docs updates needed?
No

### DIFF
--- a/charts/temporal/templates/server-deployment.yaml
+++ b/charts/temporal/templates/server-deployment.yaml
@@ -59,6 +59,8 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: status.podIP
+            - name: SERVICES
+              value: {{ $service }}
             - name: TEMPORAL_SERVICES
               value: {{ $service }}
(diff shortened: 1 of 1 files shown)


PurseChicken:
This was replaced with https://github.com/temporalio/helm-charts/pull/857 due to the overhaul of the upstream chart.

PR #762, opened 2025-10-20
Adding default image check

PhillypHenning:
## What was changed
Adding an image check test on trigger pull_request, this builds the manifest, pulls the images+tags being used and checks that they are available.

## Why?
0.68.0 had a breaking change that lasted a weekend, this should catch this and prevent it happening again.


robholland:
We have CI now which will check that the chart can install with various values files, that should catch this case.

PR #771, opened 2025-11-05
Allow configuring the setup job's ttlSecondsAfterFinished value

schiermike:
## What was changed

`schema.setup.ttlSecondsAfterFinished` is now exposed via `values.yaml`.

## Why?

We run a temporal deployed via ArgoCD. After the current default value of ttlSecondsAfterFinished (1 day), the schema job gets removed and ArgoCD then marks the project as out-of sync. We would like to keep the finished schema job indefinitely.

## Checklist

1. How was this tested:

> Tested locally by running the helm chart.

2. Any docs updates needed

> None that I am aware of.


robholland:
Done in #846

Issue #667, opened 2025-03-13
[Feature Request] Allow setting "insecure" option for curls made by store setup initContainer


geckiss:
### Is your feature request related to a problem? Please describe.

We are trying to connect Temporal to ES as visibility storage, however, the ES endpoint requires HTTPS. Setting `scheme` to `https` (as shown below) is available. However, initContainer `setup-{{ $store }}-store` in `server-job.yaml` template will fail due to `curl: (60) SSL certificate problem: unable to get local issuer certificate` if we are using certificates issued by custom CA. I'd like to ignore the certificate - one use case might be that I'm just testing if we can deploy the Temporal instance to a new environment and connect to ES from within the new environment.

### Describe the solution you'd like

I'm not sure if it's the best solution, but it would be nice if we had an `insecure` option in the valuesfile. An example:

```yaml
elasticsearch:
    enabled: false
    external: true
    ...
    scheme: "https"
    # Set to true if you would like to, for example, disable certificate verification
    insecure: false
    ...
```

This value would be used in the server-job.yaml, like this:

```yaml
- name: setup-{{ $store }}-store
          image: "{{ $.Values.admintools.image.repository }}:{{ $.Values.admintools.image.tag }}"
          ...
            {{- else if eq $driver "elasticsearch" }}
          command: ['sh', '-c']
          args:
            - 'curl {{ if $.Values.elasticsearch.insecure -}}--insecure {{ end -}} -X PUT --fail --user "$ES_USER:$ES_PWD" $ES_SCHEME://$ES_HOST:$ES_PORT/_template/temporal_visibility_v1_template -H "Content-Type: application/json" --data-binary "@schema/elasticsearch/visibility/index_template_$ES_VERSION.json" 2>&1 &&
              curl {{ if $.Values.elasticsearch.insecure -}}--insecure {{ end -}} --head --fail --user "$ES_USER:$ES_PWD" $ES_SCHEME://$ES_HOST:$ES_PORT/$ES_VISIBILITY_INDEX 2>&1 ||
              curl {{ if $.Values.elasticsearch.insecure -}}--insecure {{ end -}} -X PUT --fail --user "$ES_USER:$ES_PWD" $ES_SCHEME://$ES_HOST:$ES_PORT/$ES_VISIBILITY_INDEX 2>&1'
            {{- end }}
```

The result would look like this if `insecure: true`:

```yaml
- name: setup-visibility-store
[...]

PR #547, opened 2024-08-13
added tls configuration for elasticsearch

0xHexE:
## What was changed
Added TLS Configuration for external elasticsearch

## Why?
If we have external elasticsearch then we should able to connect to elasticsearch

## Checklist
- [ ] Setup external elasticsearch using elastic operator example
- [ ] Setup 

1. Closes #447 

2. How was this tested:

- Setup of elasticsearch 
- Install helm chart with this values

```yaml
elasticsearch:
  enabled: false
  external: true
  host: "elasticsearch-es-http"
  port: "9200"
  version: "v7"
  scheme: "https"
  logLevel: "error"
  username: "#USER#"
  password: "#PASSWORD#"
  tls:
    enabled: true
    secretName: "elasticsearch-es-remote-ca"
    ca: "ca.crt"
```

PR #774, opened 2025-11-10
Allow unsecure es check in init

adamovittorio:
## What was changed
Added `--insecure` to the `check-elasticsearch-index` init containers curl commands.

## Why?
Because at the moment when configuring the Temporal chart to connect to Elasticsearch in HTTPS, these checks fail, preventing the deployment from succeeding. Considering that's just a check to see if Elasticsearch is online and the visibility index exists, seems to me that adding the `--insecure` flag is the path of least resistance without overcomplicating things. 

## Checklist

1. Closes 

Partially solves #667 

2. How was this tested:

We have tried this by applying the change to the rendered manifest. To replicate the problem, you must connect to an Elasticsearch instance with HTTPS enabled and the certificates properly configured. 

3. Any docs updates needed?

I added an example in the `values.elasticsearch.yaml`.

### DIFF
--- a/charts/temporal/values/values.elasticsearch.yaml
+++ b/charts/temporal/values/values.elasticsearch.yaml
@@ -1,8 +1,26 @@
+# server:
+#   # Additional volumes and volume mounts for Elasticsearch TLS certificates
+#   additionalVolumes:
+#     - name: temporal-es-tls
+#       secret:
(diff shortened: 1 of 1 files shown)


robholland:
Closing in favour of the more complete solution in #547

Issue #775, opened 2025-11-13
[Feature Request] allow minReadySeconds for deployments


qs-synth:
### Is your feature request related to a problem? Please describe.

When history pods are replaced during deployment, shard rebalancing occurs. Pods are not given enough time during replacement to complete shard rebalancing before another pod is removed. This is due to the lack of a readiness probe for history pods to let the orchestrator know to pause pod replacement until rebalancing is complete for the latest pod.

A similar behavior occurs during matching pods replacement with task queue handling. Both situations lead to spikes in latency and possibly `service_unavailable` errors for history and matching services during deployment.

### Describe the solution you'd like

In the short term. allowing `minReadySeconds` to be set per deployment would solve this problem for most users so they can tune their deployment to allow the needed time between pods.

In the long term, readiness probes per-service would be ideal so that each pod can report its status upon deployment instead of relying on a static delay. Conservative users could continue to use `minReadySeconds` even after readiness probes are in place for an extra delay.

PR #776, opened 2025-11-13
feat: add minReadySeconds option

qs-synth:
## What was changed

Added minReadySeconds option for each deployment

## Why?

https://github.com/temporalio/helm-charts/issues/775

## Checklist

1. Closes https://github.com/temporalio/helm-charts/issues/775

2. How was this tested:

Setting minReadySeconds and then deploying. Pods will come up more gradually, giving them time for startup activities.

3. Any docs updates needed?

This is a standard feature in the latest Deployment spec. No need for special docs


quangngotan95:
I opened a different PR here https://github.com/temporalio/helm-charts/pull/733
I'm happy to close mine and support this as well

qs-synth:
> I opened a different PR here #733 I'm happy to close mine and support this as well

I would appreciate that! I did not see your PR before filing this one. I think this setup fits the standard of enabling separate values by service that @PhillypHenning is [suggesting on the other change](https://github.com/temporalio/helm-charts/pull/733#pullrequestreview-3190580791).

robholland:
We should probably default to a server.minReadySeconds option for the server services like other fields do. Please can you add tests?

qs-synth:
> We should probably default to a server.minReadySeconds option for the server services like other fields do. Please can you add tests?

@robholland , thanks for the feedback. Does the latest change fit your suggestion?

PR #786, opened 2025-11-21
Update Config Templating

chaptersix:
## What was changed
add required comment to enable templating and updated the `env` syntax to use `sprig`. Existing syntax if for `dockerize`.

## Why?
The config yaml includes values than are intended to be read from the environment by `sprig`. The server now requires that `# enable-template` comment to allow this.

## How was this tested:
running the helm chart with the new docker images. Everything starts successfully after this comment is added. 

**Breaking change**: existing server versions will not be able to parse this config-map


robholland:
There are multiple ways in which the new template syntax is different from `dockerize`, it just happens that `env ".."` vs `.Env.` is the only relevant issue in the helm charts config. I don't think supporting `.Env` is a good idea, we have the `enable-template` system precisely so people can choose when they opt-in to the new format.

chaptersix:
> There are multiple ways in which the new template syntax is different from `dockerize`, it just happens that `env ".."` vs `.Env.` is the only relevant issue in the helm charts config. I don't think supporting `.Env` is a good idea, we have the `enable-template` system precisely so people can choose when they opt-in to the new format.

After some internal conversation we are going with `sprig` only support.

robholland:
I'd like to introduce this into the alpha 1.0 builds rather than the 0.x range so all breaking changes occur there. Please switch the PR to target the `v1` branch rather than `main`

charts/temporal/templates/server-configmap.yaml
```diff
@@ -83,14 +84,15 @@ data:
       membership:
         name: temporal
         maxJoinDuration: 30s
-        broadcastAddress: {{ `{{ default .Env.POD_IP "0.0.0.0" }}` }}
+        # TODO: validate broadcast address can work with 0.0.0.0
+        broadcastAddress: {{ `{{ env "POD_IP" | default "" }}` }}
```
robholland: Presumably we can remove the `default` here?

chaptersix:
moving to #794

Issue #737, opened 2025-08-15
[Bug] schema job fails


till:
### What are you really trying to do?

Trying to setup temporal using this helm chart.

### Describe the bug

The schema job fails to complete — it says it cannot connect to the database.

### Minimal Reproduction

You can follow the instructions in the README in this repository for deploying Temporal with your own PostgreSQL database. Walk through the steps with the sql tool (all successful) and then helm install the latest version of the chart with the postgresql values (adjusted).

### Environment/Versions

- OS and processor: helm
- Temporal **Chart** Version: 0.65.0
- Are you using Docker or Kubernetes or building Temporal from source? k8s / AKS

### Additional context

If I disable the tool, I can access the UI:

```yaml
schema:
  createDatabase:
    enabled: false
  setup:
    enabled: false
  update:
    enabled: false
```

luiscosta18:
Hello @till 

Please have a look on https://github.com/temporalio/helm-charts/pull/686.
I'm using postgresql helm chart as a dependency and the schema job completes without any issue.
If you are using an external database, please provide your values.yaml with the database connection variables and the error returned by the job pod.

till:
@luiscosta18 hey, thanks for responding.

Can you explain how the PR addresses my issue?

I probably didn't explain it right:

My issue is not setting up PostgreSQL. My issue is that the current documentation contains all these steps to build a schema tool etc. to setup the database and run initial migrations when apparently the helm chart already comes with a container/setup job?

I followed the steps in the readme which led me here.

As I stated, if I disable the schema job, the UI seems to work? Before the UI was unavailable (error 500) and the Job setup by the helm chart failed.

till:
Oh, there may also be additional problems that the schema job is unable to tell that the system was initialized?

till:
Btw, I am personally fine with the manual setup steps. I just think then there should be something in the readme to tell people to disable the settings I did.

luiscosta18:
Hello @till,

The purpose of schema job is to bootstrap temporal and temporal visibility databases schemas automatically. There is no need to set  setup.enabled: false +  update. enabled: false and run the migrations manually.
If you are getting the error "cannot connect to the database", something may be wrong with your values.yaml. Can you please provide the error returned by the job pod and your values.yaml?

till:
![image](https://github.com/user-attachments/assets/1a646064-ca21-4114-94f5-60ffb912bd99)

From your readme **on this repository**, the instructions. This is what I followed.

[...]

Issue #709, opened 2025-06-02
[Feature Request] Support ArgoCD deployments


stephen-harris:
### Is your feature request related to a problem? Please describe.

We deploy our manifests using ArgoCD and it fails when we make a change to the Job's pod specification (e.g. the secrets being mounted, the environment variables set or image version) as by default ArgoCD uses apply and due to how Temporal's helm chart configures the schema job name, the job name is static and not configurable.

Temoporal's helm template assumes it is being deployed in the context of a helm release with the job name based on the release revision: (https://github.com/temporalio/helm-charts/blob/c05acfe50c33f18942b4c3aaad748a2866323a37/charts/temporal/templates/server-job.yaml#L13), but outside of that context the job name remains at '-1'. As jobs are immutable subsequent changes to the job template will fail.

### Describe the solution you'd like

There are two options I can think of:

 - Allow annotations to be added to the job spec. This allows us to instruct ArgoCD via annotations to replace the resource rather than apply updates. I've raised a PR for this option here: https://github.com/temporalio/helm-charts/pull/665
 - Alternatively, make the job name configurable / over-ridable. This is slightly less ideal, but if we can manually 'bump' the job name, that will allow us to work around this issue.

cedson:
I'm running into this issue as well - have to manually drop the schema-1 job each time I need to upgrade, etc.

Other possible considerations:
- Potentially add usage of helm-hooks that define it as pre-install or pre-sync (ArgoCD runs all at once, fyi), and another to remove the job pre-sync;
- explore usage of 'generateName' ?

jmooo:
as a hack workaround could we use a pre-sync hook that deletes the existing schema job before the sync tries to re-create it? I believe this would be idempotent since it doesn't actually touch anything inside the db

```
apiVersion: batch/v1
kind: Job
metadata:
  generateName: temporal-schema-cleanup-
  namespace: temporal
  annotations:
    argocd.argoproj.io/hook: PreSync
    argocd.argoproj.io/hook-delete-policy: HookSucceeded
spec:
  ttlSecondsAfterFinished: 120
  template:
    spec:
      serviceAccountName: argocd-temporal-cleanup  # Needs RBAC
      restartPolicy: Never
      containers:
      - name: cleanup
        image: bitnami/kubectl:latest
        command:
        - /bin/sh
        - -c
        - |
          kubectl delete job -n temporal \
            -l app.kubernetes.io/component=schema \
            --ignore-not-found=true
[...]

Issue #636, opened 2025-01-27
Inconsistent diff using Helm diff plugin


edsonmarquezani:
Since [this change](https://github.com/temporalio/helm-charts/pull/564), I've been getting a constant diff every time I run [Helm diff](https://github.com/databus23/helm-diff). It always tells it's going to delete the current version and replace it with a job named simply `temporal-schema-1`.

Example:

```diff
temporal, temporal-schema-34, Job (batch) has been removed:
- apiVersion: batch/v1
- kind: Job
- metadata:
-   labels:
-     app.kubernetes.io/component: database
-     app.kubernetes.io/instance: temporal
-     app.kubernetes.io/managed-by: Helm
-     app.kubernetes.io/name: temporal
-     app.kubernetes.io/part-of: temporal
-     app.kubernetes.io/version: 1.26.2
-     helm.sh/chart: temporal-0.54.0
-   name: temporal-schema-34
- spec:
-   backoffLimit: 100
[...]

temporal, temporal-schema-1, Job (batch) has been added:
+ apiVersion: batch/v1
+ kind: Job
+ metadata:
+   labels:
+     app.kubernetes.io/component: database
+     app.kubernetes.io/instance: temporal
+     app.kubernetes.io/managed-by: Helm
+     app.kubernetes.io/name: temporal
+     app.kubernetes.io/part-of: temporal
+     app.kubernetes.io/version: 1.26.2
+     helm.sh/chart: temporal-0.54.0
+   name: temporal-schema-1
+ spec:
[...]
```

It seems like the diff plugin isn't able to look up `.Release.Revision` from the actual current release. The weirdest part is: if I apply it anyway, the job will in fact be replaced, but instead of creating the `temporal-schema-1` as shown, the release preffix will be bumped as expected. Next time I run `diff`, the same thing will happen, except the job's name has been bumped in 1 unit.

I've tried using `HELM_DIFF_THREE_WAY_MERGE`, ` HELM_DIFF_USE_INSECURE_SERVER_SIDE_DRY_RUN` and `HELM_DIFF_USE_UPGRADE_DRY_RUN`. No success.

I'm not sure using `.Release.Revision` in the job name is the best solution. Maybe the `.Chart.Version` instead? Or at least providing a backwards compatibility option to keep the job name always the same.

If anyone knows a way to work around this issue currently, I would appreciate too.

edsonmarquezani:
@robholland tagging you here because you're the author of the original change.

[...]

PR #792, opened 2025-11-30
Create a unique setup/update Job name per release version

### LINKED ISSUE #554: [Bug] `helm upgrade` does not work if `schema.enabled` is true

jorhett:
## Why?

The helm chart currently creates a job to setup and update the schema named `temporal-schema-{{Helm chart version}}` since #564 which is often just `temporal-schema-1`. The lifetime of this chart is 1 day, which has two separate impacts:

* The chart must be manually deleted before an upgrade can occur 
  * #554
  * #735
* ArgoCD shows OutOfSync / Helm diff after Job expires
  * #636 
  * #709
  * #743
* Instructions say a user must run the upgrade, but the chart does it for you if `schema.update` is enabled
  * #737

## What was changed

1. Give the schema update job and pod a name including the major-minor version of Temporal. This will allow multiple minor version upgrades to succeed without having to manually remove the job.

2. Remove the job removal TTL the value of which was debatable, but regardless of value will show sync issues with  Helm diff and ArgoCD sync status. As the use of Helm hooks was explicitly removed in #522 to resolve issues with `--wait`, this provides another way to retain diff/sync. This also means that job update logs are retained and available for life of the helm release.

3. Updated the README to make it clear the manual update steps aren't required if update is enabled.

## Checklist

1. Closes #554 #735 #636 #709 #737

3. How was this tested: it's only a name change, helm upgrade did it all 😁 

5. Any docs updates needed?

Updates to the README are included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -529,7 +529,10 @@ helm uninstall temporaltest
 
 ## Upgrading
 
-To upgrade your cluster, upgrade your database schema (if the release includes schema changes), and then use `helm upgrade` command to perform a rolling upgrade of your installation.
+To upgrade your cluster:
+
+* Upgrade your database schema (if `schema.update.enabled` is `false`)
+* Run `helm upgrade` command to perform a rolling upgrade of your installation
 
 Note:
 * Not supported: running newer binaries with an older schema.
@@ -538,6 +541,8 @@ Note:
 Example:
 
 ### Upgrade Schema
+
+**These steps are only required if `schema.update.enabled` is `false`**. If schema update is enabled the Helm chart will do this for you.
 
 Here are examples of commands you can use to upgrade the "default" schema in your "bring your own" Cassandra database.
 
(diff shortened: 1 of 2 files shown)


robholland:
Please could you update this now that v1 has been merged to master. Please add a test, and consider how this might work given we're currently doing rc releases? Might need to have support for pre/alpha/rc style stuff in the schema job name as well.

jorhett:
> consider how this might work given we're currently doing rc releases? 

It's kind of odd to respond to a submission and say "please go solve this new problem specific to our releases" -- especially when I know nothing of your versioning or release strategy 😁

I've demonstrated a useful way to solve this problem and minimize problems that lots of people were experiencing in many different ways. I would think details like adapting it to changes in your versioning strategy / etc are important to you, but orthogonal to the value provided by this PR.  Please take this answer and implement it however you feel is best.

jorhett:
Also keep in mind that AFAICT the "changes" required would be altering a `printf()` call to properly represent your versions.  Trivial (<10 chars?) change for someone who knows your versioning strategy, and a large investment of research by someone who doesn't.

I'm not in any way trying to suggest that version handling isn't important, I'm saying an engineer who knows your versioning can do it between breaths 😉

robholland:
Full release name is now used for the job via #846

PR #803, opened 2025-12-15
Add dual visibility support

mihaelabalas84:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


robholland:
We have just merged the v1 branch to main which has simplified a lot of the helpers and templates. If you're able to adjust the PR to the new structure (which I think will be much cleaner than the previous) then we'd love to get this feature merged.

PR #813, opened 2025-12-19
feat: add test pod annotations support

Tomlord1122:
## What was changed                                                                                 
                                                                                                      
Added support for custom pod annotations on the Helm test pod by:                                   
  1. Updating `templates/tests/test.yaml` to render values from `test.podAnnotations`                 
  2. Adding `test.podAnnotations: {}` configuration option to `values.yaml`                           
                                                                                                      
  This allows users to specify custom annotations (e.g., Istio sidecar configuration) on the test pod without modifying the template directly.
                                                                                                      
  ## Why?                                                                                             
                                                                                                      
  In Istio-enabled environments, the test pod (`helm test`) hangs indefinitely because:               
  - The test pod completes and exits                                                                  
  - But the Istio sidecar doesn't terminate                                                           
  - Causing the test to timeout and fail                                                              
                                                                                                      
  Users can now disable sidecar injection for the test pod:                                           
                                                                                                      
  ```yaml                                                                                             
  test:                                                                                               
    podAnnotations:                                                                                   
      sidecar.istio.io/inject: "false"                                                                
 ```                                                                                                      
  This aligns with the existing annotation support in other Temporal components (server, web, admintools, schema job) for consistent user experience.
                                                                                                      
##  Checklist                                                                                           
How was this tested:                                                                             
1. Verified helm lint passes with no errors                                                        
2. Verified helm template renders all 19 resources without validation errors                       
3. Template syntax checked for correct YAML indentation


robholland:
Looks good, but please add a test https://github.com/temporalio/helm-charts/tree/main/charts/temporal/tests

Tomlord1122:
> Looks good, but please add a test https://github.com/temporalio/helm-charts/tree/main/charts/temporal/tests

Thanks! I add it in this commit [b31e65b](https://github.com/temporalio/helm-charts/pull/813/commits/b31e65bc5af2fe6e54773ef0b18d816cdd93199b). 
I found that if we put `test.yaml` in `test/`. `helm-unittest` cannot recognize templates in subdirectories like `templates/tests/` due to `helmignore`. I moved `tests/test.yaml` to `templates/test.yaml` to resolve the issue.

robholland (approved):

PR #815, opened 2025-12-19
Add support for adding custom annotations to "schema" job definition

k-ikigai:
## What was changed
Added support for defining custom job annotations

## Why?
When using temporal helm chart together with ArgoCD job cannot be recreated.  
It is happening because ArgoCD is not using helm to install charts but it just renders them using helm template.
That has repercussion of job name to stay the same all the time (Revision is always 1): 
```
name: {{ include "temporal.componentname" (list $ (printf "schema-%d" .Release.Revision | replace "." "-")) }}
```

Adding a possiblility to add custom annotations allows us to define this one so that job is able to redeploy:
argocd.argoproj.io/sync-options: Replace=true,Force=true

## Checklist

1. Closes 
N/A

2. How was this tested:
I extended tests to check that newly added jobAnnotations works. I also covered podAnnotations and podLabels in that test case since it was missing. I did run tests using `helm unittest .` and they passed correctly:
image

I also did render it locally. I created values.annotations.yaml with this content:
```
schema:
  jobAnnotations:
    custom-job-annotation: abc
  podAnnotations:
    custom-pod-annotation: def
  podLabels:
    custom-pod-label: ghi

```

and using that command:
```
helm template . -f values/values.postgresql.yaml -f values/values.annotations.yaml
```

I confirmed that it renders correctly and annotations and labels are present:
```
# Source: temporal/templates/server-job.yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: release-name-temporal-schema-1
  labels:
    app.kubernetes.io/component: database
    app.kubernetes.io/name: temporal
    helm.sh/chart: temporal-1.0.0-rc.1
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/version: "1.29.1"
    app.kubernetes.io/part-of: temporal
  annotations:
    custom-job-annotation: abc
spec:
  backoffLimit: 100
  template:
    metadata:
      name: release-name-temporal-schema-1
      labels:
        app.kubernetes.io/component: database
        app.kubernetes.io/name: temporal
        helm.sh/chart: temporal-1.0.0-rc.1
        app.kubernetes.io/managed-by: Helm
        app.kubernetes.io/instance: release-name
        app.kubernetes.io/version: "1.29.1"
        app.kubernetes.io/part-of: temporal
        custom-pod-label: ghi
      annotations:
        custom-pod-annotation: def
```

3. Any docs updates needed?
N/A

### DIFF
--- a/charts/temporal/templates/server-job.yaml
+++ b/charts/temporal/templates/server-job.yaml
@@ -4,6 +4,10 @@ metadata:
   name: {{ include "temporal.componentname" (list $ (printf "schema-%d" .Release.Revision | replace "." "-")) }}
   labels:
     {{- include "temporal.resourceLabels" (list $ "database" "") | nindent 4 }}
+  {{- with $.Values.schema.jobAnnotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
 spec:
   backoffLimit: {{ $.Values.schema.backoffLimit }}
   ttlSecondsAfterFinished: 86400
--- a/charts/temporal/tests/server_job_test.yaml
+++ b/charts/temporal/tests/server_job_test.yaml
@@ -129,4 +129,36 @@ tests:
       - equal:
           path: spec.template.spec.initContainers[*].resources.limits.memory
           value: 16Gi
-
+  - it: includes custom annotations and labels for pod and job
+    set:
+      server:
+        config:
+          persistence:
+            datastores:
+              default:
+                sql:
+                  pluginName: mysql8
+                  connectAddr: "temporal-persistence:3306"
+                  databaseName: temporal
+              visibility:
+                sql:
+                  pluginName: mysql8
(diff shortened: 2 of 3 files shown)


k-ikigai:
Hey @robholland I created change that is based on the newest sources. I also added test case for my change as well as the other annotations in that job.

This is follow up based on discussion in that PR (it's not mine): https://github.com/temporalio/helm-charts/pull/665

robholland:
I don't think this is the best solution to the problem you mentioned, I prefer

robholland:
I absolutely did not hit Close! Sorry, what I was trying to say, I don't think this is the right solution to the problem, I prefer https://github.com/temporalio/helm-charts/pull/792 which I will look at. But this is a good thing to have anyway.

robholland (approved):

PR #818, opened 2025-12-24
Enable metrics annotations for matching service by default

ntnj:
## What was changed

https://github.com/temporalio/helm-charts/commit/f9fcbdf6819410e388421a2a8fb1ff363b4d5f83 but missed setting it to true for matching service requiring users to override it.

## Why?

To not have to override it separately to add metrics annotations.

## Checklist

1. Closes  N/A

2. How was this tested: N/A

### DIFF
--- a/charts/temporal/values.yaml
+++ b/charts/temporal/values.yaml
@@ -329,7 +329,7 @@ server:
       membershipAppProtocol: tcp
     metrics:
       annotations:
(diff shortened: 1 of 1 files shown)


robholland:
@ntnj Please can you sign the CLA?

robholland (approved):

PR #820, opened 2025-12-30
feat: generate TLS certificates using cert-manager

luiscosta18:
**What was changed**
Add certificates.yaml + secret.yaml to templates

**Why?**
generate certificates using cert-manager

**How was this tested**
Deployed [cert-manager helm-chart](https://github.com/cert-manager/cert-manager/tree/master/deploy)
Deployed temporal helm chart, 0.57.0, using temporal/charts/temporal/tests/certificates.yaml as values.yaml, with my own tls.crt + tls.key
Cert-manager will:

Use the temporal-issuer (which refers to the CA).
Ask the CA (from the tls-certs secret) to sign a new certificate.
Create the temporal-tls-certs secret in the same namespace — this secret will contain:
tls.crt: the signed certificate
tls.key: the private key
Optionally, ca.crt: the CA certificate (if configured)
➜  ~ kubectl get secret -n temporal

NAME                          TYPE                   DATA       AGE
temporal-tls-certs            kubernetes.io/tls      3          3h38m
tls-certs                     Opaque                 2          153m

➜  ~ kubectl get certificate  -n temporal

NAME                READY          SECRET                       AGE
temporal-cert       True           temporal-tls-certs           5h35m

➜  ~ kubectl get certificaterequests  -n temporal-data

NAME              APPROVED   DENIED   READY   ISSUER            REQUESTER                                                                      AGE
temporal-cert-1   True                True    temporal-issuer   system:serviceaccount:cert-manager:test-cert-manager  5h


luiscosta18:
Hello @robholland @tomwheeler @PhillypHenning

Please review this PR. Thanks.

luiscosta18:
> Hello @robholland @tomwheeler @PhillypHenning
> 
> Please review this PR. Thanks.

Hello @robholland @tomwheeler @PhillypHenning 
Looking forward to have your feedback on this 🙂

charts/temporal/ci/certificates.yaml
```diff
@@ -0,0 +1,90 @@
+server:
+  additionalVolumeMounts:
```
robholland, line 2: Can we automate these volumes/volumeMounts via helpers so that the user doesn't need to include them in their values? They are all boiler plate as the variable names and locations are all determined by Temporal itself. We can use a `server.tls` or similar values structure that mixes in the relevant volumes/mounts alongside any `additional*` the user has set. Likewise `web.tls`.

charts/temporal/ci/certificates.yaml
```diff
@@ -0,0 +1,90 @@
...
+      mountPath: /etc/temporal/tls
+  additionalVolumes:
+    - name: tls-certs
+      secret:
+        secretName: tls-certs
+    - name: temporal-tls-certs
+      secret:
+        secretName: temporal-tls-certs
+  config:
+    tls:
```
robholland, line 15: We can build `config.tls` automatically based on interpreting a more concise `server.tls` stanza also. Most of this is also boiler plate, just not sure about `requireClientAuth` and `serverName` for example. I defer to your knowledge of what should be customisable there.

charts/temporal/ci/certificates.yaml
```diff
@@ -0,0 +1,90 @@
...
+      value: /etc/temporal/tls/tls.key
+frontend:
+  service:
+    enabled: true
+  ingress:
+    enabled: true
+    className: ""
+    hosts:
+      - ""
+additionalSecrets:
```
robholland, line 66: Are these basically placeholders that `cert-manager` updates?

charts/temporal/ci/certificates.yaml
```diff
@@ -0,0 +1,90 @@
...
+  - name: tls-certs
+    value:
+      tls.crt: |
+        -----BEGIN CERTIFICATE-----
+        -----END CERTIFICATE-----
+      tls.key: |
+        -----BEGIN EC PRIVATE KEY-----
+        -----END EC PRIVATE KEY-----
+certificates:
```
robholland, line 75: Ah, maybe we add the volumes/mounts based on this block? Not sure if need to have `server.` and `web.` variants though, would it be valid to want cert manager for one but not the other?

charts/temporal/templates/secret.yaml
```diff
@@ -0,0 +1,13 @@
+{{- if .Values.additionalSecrets }}
```
robholland: Let's not add an additionalSecrets concept here, but rather have secrets specific to TLS. I think there is actually a specific kind of secret in k8s for certificates?

robholland (changes requested):

charts/temporal/templates/secret.yaml
luiscosta18: yes

robholland:
Closing in favour of #954, inspired by your work. Thanks for the time you spent on this!

PR #822, opened 2025-12-31
Fix templating issue for persistence with existingSecret

### LINKED ISSUE #821: [Bug] Password field not injected into config when using `existingSecret` without explicit `password` field

Configure postgres as default store with chart version `temporal-1.0.0-rc.1`

### Describe the bug

When configuring SQL persistence stores using `existingSecret` and `secretKey` to reference Kubernetes secrets, the chart correctly creates environment variables from the secret but fails to inject the `password` field into the Temporal server configuration template. This causes authentication failures with errors like:

```bash
failed SASL auth: FATAL: password authentication failed for user "postgres" (SQLSTATE 28P01)
```

### Configuration

```yaml
server:
  config:
    persistence:
      datastores:
        default:
          sql:
            user: postgres
            existingSecret: temporal-psql-creds
            secretKey: password
            # No explicit password field
```

### Expected Behavior

The chart should automatically inject a `password` field in the rendered config template that references the environment variable created from the secret:

```yaml
datastores:
  default:
    sql:
      user: postgres
      password: {{ env "TEMPORAL_DEFAULT_STORE_PASSWORD" | quote }}
```

### Actual Behavior

The rendered config (in the `temporal-config` ConfigMap) contains:

```yaml
datastores:
  default:
    sql:
      user: postgres
      # no password field at all
```

However, the environment variable `TEMPORAL_DEFAULT_STORE_PASSWORD` is correctly created in the pods and contains the correct password value. If we set password within the values.yaml in addition to the existingSecret/secretKey all the services successfuly start.

amithkk:
## What was changed

Modified the temporal.persistence.filterConfig function in `templates/_helpers.tpl` to check for both password and `existingSecret` fields when determining whether to inject password environment variable references into the server configuration.

## Why?

The chart was only injecting the password field into the rendered config template if users explicitly specified a password field in their values.yaml. This meant that when users relied solely on `existingSecret` and `secretKey` to reference Kubernetes secrets (as documented and intended), the password field was completely missing from the configuration template, causing authentication failures.

## Checklist

1. Closes  #821 

2. How was this tested:
By successfully deploying the modified chart onto a kubernetes cluster, following the postgres example with `existingSecret` and `secretKey` only

3. Any docs updates needed?
No

### DIFF
--- a/charts/temporal/tests/server_configmap_test.yaml
+++ b/charts/temporal/tests/server_configmap_test.yaml
@@ -94,3 +94,43 @@ tests:
       - matchRegex:
           path: data['config_template.yaml']
           pattern: "connectAddr: postgres.example.com:5432"
+
+  - it: injects password from existingSecret without explicit password field
+    set:
+      server:
+        enabled: true
+        config:
+          persistence:
+            defaultStore: default
+            visibilityStore: visibility
+            numHistoryShards: 512
+            datastores:
+              default:
+                sql:
+                  pluginName: postgres12_pgx
+                  driverName: postgres12_pgx
+                  databaseName: temporal
+                  connectAddr: "postgres.example.com:5432"
+                  user: temporal_user
+                  existingSecret: my-postgres-secret
+                  secretKey: postgresPassword
+              visibility:
+                sql:
+                  pluginName: postgres12_pgx
+                  driverName: postgres12_pgx
+                  databaseName: temporal_visibility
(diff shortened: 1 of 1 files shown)


robholland:
Thanks for catching this, I'll add a ci test to ensure we cover this case. Please could you add a unit test for this?

amithkk:
@robholland Sure! done.

PR #828, opened 2026-01-10
Fix sql port in admin tools when not provided

efiShtain:
## What was changed
This change try to set the correct SQL_PORT when using sql plugin and providing an address like my-db.endpoint.com without a port.
Current implementation sets the port to my-db.endpoint.com since it tries to use the last part after : and there is only one part

## Why?
default port can be assumed and it is better than invalid port which is current implementation

2. How was this tested:
manually
- without port, it takes the default store's port (5432 for postgres and 3306 for mysql)
- with port it takes the correct one

3. Any docs updates needed?
not sure

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,8 @@ server:
       datastores:
         default:
           sql:
+            createDatabase: true
+            manageSchema: true
             pluginName: mysql8  # or postgres12, postgres12_pgx
             driverName: mysql8
             databaseName: temporal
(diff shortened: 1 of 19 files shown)


robholland:
I'd prefer we fail the chart if they don't provide a port number, rather than us providing default ports. It is not unreasonable to require a port number.

efiShtain:
@robholland 
I tend to agree, wanted to be less disruptive to current users

fixed it to fail if port isn't provided

let me know if any other changes are required

robholland (changes requested):

PR #830, opened 2026-01-13
feat: Add activeDeadlineSeconds for server job

### LINKED ISSUE #829: [Feature Request] Add activeDeadlineSeconds for server job
https://kubernetes.io/docs/concepts/workloads/controllers/job/
Please add activeDeadlineSeconds for server job
1. Prevents Stuck or Hanging Jobs
2. Ensures Better Resource Management
3. Prevents CronJobs from Getting Stuck Due to Too Many Missed Schedules
4. Improves Cluster Stability

minivolk:
1. Closes #829

2. How was this tested:
I run helm template locally with my changes

3. Any docs updates needed?
Only for values.yaml

### DIFF
--- a/charts/temporal/values.yaml
+++ b/charts/temporal/values.yaml
@@ -491,6 +491,7 @@ web:
   podDisruptionBudget: {}
 schema:
   backoffLimit: 100
+  activeDeadlineSeconds: 500
   jobAnnotations: {}
   podAnnotations: {}
   podLabels: {}
--- a/charts/temporal/templates/server-job.yaml
+++ b/charts/temporal/templates/server-job.yaml
@@ -10,6 +10,7 @@ metadata:
   {{- end }}
 spec:
   backoffLimit: {{ $.Values.schema.backoffLimit }}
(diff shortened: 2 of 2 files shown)


charts/temporal/values.yaml
```diff
@@ -491,6 +491,7 @@ web:
   podDisruptionBudget: {}
 schema:
   backoffLimit: 100
+  activeDeadlineSeconds: 500
```
robholland: Please leave a null default so we don't change behaviour unless a user needs it.

robholland (changes requested):

charts/temporal/values.yaml
minivolk: > Please leave a null default so we don't change behaviour unless a user needs it.

```diff
 schema:
   backoffLimit: 100
-  activeDeadlineSeconds: 500
+  activeDeadlineSeconds: null
   ttlSecondsAfterFinished: 86400
```

Done

robholland (approved):

PR #832, opened 2026-01-16
docs: comment createDatabase and manageSchema

ArkShocer:
## What was changed
Documented on how to disable DB creation and Schema management

## Why?
I did not find any up-to-date documentation on how to disable db creation for the rc. I only found something after looking through the templates job. 
https://github.com/temporalio/helm-charts/blob/4f47773f69ab237615c48e26c4f8626898558550/charts/temporal/templates/server-job.yaml#L34
https://github.com/temporalio/helm-charts/blob/4f47773f69ab237615c48e26c4f8626898558550/charts/temporal/templates/server-job.yaml#L61

## Checklist

1. Closes 
https://github.com/temporalio/helm-charts/issues/440
2. How was this tested:
Set the `createDatabase: false` parameter on an existing database and it went through

3. Any docs updates needed?
Did not find anything on the `docs.temporal.io` page when searching for `createDatabase` or similar searches. Maybe add some information here? https://docs.temporal.io/self-hosted-guide/visibility#postgresql

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,8 @@ server:
       datastores:
         default:
           sql:
+            createDatabase: true
+            manageSchema: true
             pluginName: mysql8  # or postgres12, postgres12_pgx
             driverName: mysql8
             databaseName: temporal
@@ -75,6 +77,8 @@ server:
             maxConnLifetime: "1h"
         visibility:
           sql:
+            createDatabase: true
+            manageSchema: true
             pluginName: mysql8
             driverName: mysql8
             databaseName: temporal_visibility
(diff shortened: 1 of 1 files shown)


robholland:
Good catch. These should be added to the README, would you mind doing that as part of this PR?

ArkShocer:
@robholland I've added them to the readme. Wouldn't it be better to add a link from the readme to the `values\values.[DATABASE].yaml` so you have one location where the best example lies? Like you have it with https://github.com/temporalio/helm-charts?tab=readme-ov-file#install-with-mysql but without the code part in the readme as it's not the same + the link to the values yaml is wrong :D

robholland:
I don't really like that people have to skip back and forward from the values file to see the blocks docs wise. I was meaning to add some hints in https://github.com/temporalio/helm-charts/?tab=readme-ov-file#persistence-configuration. I'll sort that in another PR though.

PR #833, opened 2026-01-17
Allow setting additional environment variables from ConfigMap too

ntnj:
## What was changed

Add an option to add environment variable from configmap, similar to existing option for secrets

## Why?

It's more ergonomical to add all environment variables at once in a configmap or a secret, instead of adding them in additionalEnv. Using ConfigMap makes it easier to use than a Secret

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


robholland:
Please add tests for deployments.

PR #834, opened 2026-01-17
Allow setting other config values in metrics

ntnj:
## What was changed

Allow setting other [metrics fields](https://github.com/temporalio/temporal/blob/main/common/metrics/config.go) in config template.

Move the configuration related metrics field from `server.metrics` to `server.config.metrics`

## Why?

Add `server.config.metrics` which is used to generate metrics section for config for consistency with other configuration related data.
Default value of prometheus configuration has been moved to values.yaml.

## Checklist

1. Closes 

2. How was this tested:

Tested the full template by printing out various combinations. Verified no diff with default values.

3. Any docs updates needed?

Might be need to mention in breaking changes section of v1.

### DIFF
--- a/charts/temporal/templates/server-configmap.yaml
+++ b/charts/temporal/templates/server-configmap.yaml
@@ -38,25 +38,11 @@ data:
       metrics:
         tags:
           type: {{ `{{ env "TEMPORAL_SERVICES" | quote }}` }}
-          {{- with (merge $server.config.metrics.tags $server.metrics.tags) }}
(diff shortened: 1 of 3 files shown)


robholland:
Please create `server.config.metrics` as you suggest, now is a good time to make such a change.

robholland:
We don't need to handle backwards compatibility here as we're heading to a new major release. Let's drop the complexity as we have the opportunity to clean up the confusing setup.

PR #835, opened 2026-01-17
Add an option to disable creating shims

ntnj:
## What was changed

Added an option to disable creation of dockerize and elasticsearch-tool shims. They're needed in <= v1.29, but can be disabled for >= v1.30.

## Why?

To avoid unnecessary volume mounts when they're not needed.

## Checklist

1. Closes 

2. How was this tested:

Compared results of `helm template .` with and without this change.

3. Any docs updates needed?


robholland:
Please add tests for these toggles.

ntnj:
Does the added test look good? I'm not sure how to test this otherwise.

PR #836, opened 2026-01-19
fix: minReadySeconds usage with deployments

tskibinski:
## What was changed
Corrected usage of `minReadySeconds` option for each deployment. 

## Why?
In the Kubernetes API, `minReadySeconds` is a field of the Deployment Specification (`DeploymentSpec`), not the Pod Specification (`PodSpec`).

https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#min-ready-seconds

It must be a direct child of the Deployment `spec`, at the same indentation level as `replicas`, `selector`, and `template`.

If it is under `spec.template`, it is effectively being passed to the Pod definition, where it is not a valid field and will likely be ignored by the Deployment controller (or cause a validation error depending on the strictness of your cluster).

## Checklist

2. How was this tested:

Compared results of `helm template .` with and without this change.

3. Any docs updates needed?

This is a standard feature in the latest Deployment spec. No need for special docs

### DIFF
--- a/charts/temporal/tests/admintools_deployment_test.yaml
+++ b/charts/temporal/tests/admintools_deployment_test.yaml
@@ -29,3 +29,11 @@ tests:
       - equal:
           path: spec.template.spec.containers[0].volumeMounts[0].name
           value: my-volume
+  - it: minReadySeconds usage
+    set:
+      admintools:
+        minReadySeconds: 10
+    asserts:
+      - equal:
+          path: spec.minReadySeconds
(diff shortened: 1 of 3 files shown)


robholland:
Thanks for catching this. Please can you update/add tests as appropriate as well?

tskibinski:
@robholland Sorry, I somehow missed them. I fixed the `server` Deployment tests and added also for `web` and `admintools` deployment.

PR #856, opened 2026-03-04
Enable conditional rendering for service definitions

### LINKED ISSUE #855: [Bug] server-service.yaml does not respect `enabled: false` for individual services

When deploying Temporal via Helm chart, I want to disable specific services (e.g., `internal-frontend`) by setting `server.<service>.enabled: false` in `values.yaml`, and expect that the corresponding Kubernetes `Service` resources are **not** created.

### Describe the bug

In the current Helm templates, the `Service` resources under `templates/server-service.yaml` are generated for all services in the list (`frontend`, `internal-frontend`, `matching`, `history`, `worker`) regardless of whether the individual service is disabled via `server.<service>.enabled: false`.

The template currently iterates over the service list but does **not** check `$serviceValues.enabled` before rendering the `Service` resource. As a result, even when a service like `internal-frontend` is disabled in `values.yaml`, its corresponding Kubernetes `Service` is still created.

**Expected behavior:**  
When `server.internal-frontend.enabled` is set to `false`, the Kubernetes `Service` for `internal-frontend` should **not** be created.

**Actual behavior:**  
The Kubernetes `Service` for `internal-frontend` is created even when `server.internal-frontend.enabled` is `false`.

**Root cause:**  
The `server-service.yaml` template is missing a conditional check (`{{- if $serviceValues.enabled }}`) before rendering each `Service` resource within the `{{- range }}` loop.

### Minimal Reproduction

1. Set `server.internal-frontend.enabled: false` in `values.yaml`
2. Run `helm template` or `helm install`
3. Observe that the `Service` resource for `internal-frontend` is still rendered/created

### Environment/Versions

- OS and processor: macOS (Apple Silicon)
- Temporal Version: \[fill in your version\]
- Deployment method: Kubernetes (Helm chart)

### Additional context

The fix involves adding `{{- if $serviceValues.enabled }}` guard inside the `{{- range }}` loop in `templates/server-service.yaml`, similar to how other templates (e.g., `server-deployment.yaml`) already handle this pattern:

```yaml
{{- range $service := (list "frontend" "internal-frontend" "matching" "history" "worker") }}
{{- $serviceValues := index $.Values.server $service }}
{{- if $serviceValues.enabled }}
# ... Service resource definition ...
{{- end }}
{{- end }}
{{- end }}
```

lake-dunamu:
## What was changed
Added `{{- if or (not (hasKey $serviceValues "enabled")) $serviceValues.enabled }}` conditional check inside the `{{- range }}` loop for the non-headless Service block in `templates/server-service.yaml`.

## Why?
The non-headless Service block for `frontend` and `internal-frontend` had no `enabled` check, so Service resources were created regardless of the `enabled` flag. The headless Service block already uses `{{- if $serviceValues.enabled }}`. This change makes the non-headless block consistent while using an opt-out pattern — services without an explicit `enabled` key are rendered by default, and only those with `enabled: false` are skipped.

## Checklist
- Closes #855
- How was this tested:
   - `helm template` with `server.internal-frontend.enabled: false` → internal-frontend Service not rendered.
   - `helm template` with `server.internal-frontend.enabled: true` → Service rendered as expected.
   - `helm template` without `enabled` key set → Service rendered by default (opt-out behavior).
   - Verified other services are unaffected.
- Any docs updates needed?
   - No.

### DIFF
--- a/charts/temporal/tests/server_service_test.yaml
+++ b/charts/temporal/tests/server_service_test.yaml
@@ -51,6 +51,40 @@ tests:
         containsDocument:
           kind: Service
           apiVersion: v1
+  - it: doesn't create frontend services when frontend is disabled
+    template: templates/server-service.yaml
+    set:
+      server:
+        frontend:
+          enabled: false
+    asserts:
+      - not: true
+        containsDocument:
+          kind: Service
+          apiVersion: v1
+          name: RELEASE-NAME-temporal-frontend
+      - not: true
+        containsDocument:
+          kind: Service
+          apiVersion: v1
+          name: RELEASE-NAME-temporal-frontend-headless
+  - it: doesn't create internal-frontend services when internal-frontend is disabled
+    template: templates/server-service.yaml
+    set:
+      server:
+        internal-frontend:
+          enabled: false
+    asserts:
+      - not: true
+        containsDocument:
+          kind: Service
+          apiVersion: v1
+          name: RELEASE-NAME-temporal-internal-frontend
+      - not: true
+        containsDocument:
+          kind: Service
+          apiVersion: v1
+          name: RELEASE-NAME-temporal-internal-frontend-headless
   - it: app protocol default in frontend service
     template: templates/server-service.yaml
     documentSelector:


lake-dunamu:
@robholland could you PTAL?

We need this fix because our security policy requires removing all endpoints for disabled services, and currently the non-headless Service is still created even with `enabled: false.`

robholland:
Looks good, please add a test to `charts/temporal/tests/server_service_test.yaml`.

lake-dunamu:
> Looks good, please add a test to charts/temporal/tests/server_service_test.yaml.

@robholland  Done. Thanks!

PR #857, opened 2026-03-06
feat: Allow the ability to configure garbage collection on schema-job.

### LINKED ISSUE #735: [Feature Request] Allow Schema Job ttlSecondsAfterFinished to be configurable

PurseChicken:
## What was changed

This PR gives the user the ability to configure schema-job garbage collection. Specifically the ability to enable or disable the garbage collection altogether as well as specify the desired time before garbage collection runs against the job if enabled.

See [Here](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/) for more information about kubernetes garbage collection on job resources.

## Why?
Currently garbage collection (.spec.ttlSecondsAfterFinished) is hard coded in the server-job manifest. This means that the job will always delete itself after the specified value (currently 86400). In some cases users may want to either configure the number of seconds before deleting the job manifest, or they may want to disable this garbage collection altogether. This PR takes care of both of those scenarios.

## Checklist

2. How was this tested:
The chart was validated using helm template --debug as well as packaging it (helm package) and using it in a deployment.

3. Any docs updates needed?
No

### DIFF
--- a/charts/temporal/templates/server-job.yaml
+++ b/charts/temporal/templates/server-job.yaml
@@ -10,7 +10,11 @@ metadata:
   {{- end }}
 spec:
   backoffLimit: {{ $.Values.schema.backoffLimit }}
+  {{- if and $.Values.schema.garbageCollection $.Values.schema.garbageCollection.enabled }}
+  ttlSecondsAfterFinished: {{ $.Values.schema.garbageCollection.ttlSeconds | default 86400 }}
+  {{- else if not $.Values.schema.garbageCollection }}
   ttlSecondsAfterFinished: 86400
+  {{- end }}
   template:
     metadata:
       name: {{ include "temporal.componentname" (list $ (printf "schema-%d" .Release.Revision | replace "." "-")) }}
--- a/charts/temporal/values.yaml
+++ b/charts/temporal/values.yaml
@@ -508,6 +508,11 @@ web:
   minReadySeconds: 0
   podDisruptionBudget: {}
 schema:
+  garbageCollection:
+    # Enable or disable the job garbage collection
+    enabled: true
+    # If enabled, time before the job is removed in seconds. Default 86400
+    ttlSeconds: 86400
   backoffLimit: 100
   jobAnnotations: {}
   podAnnotations: {}


robholland:
After #846 you can set `ttlSecondsAfterFinished: null` in values to turn off gc. https://github.com/temporalio/helm-charts/pull/846/changes#diff-144f01bd3f0800a758f91c5e3e8a24f2177d14216e25918d924c5f3eb5d1a467R526

PurseChicken:
@robholland 

We should not rely on ttlSecondsAfterFinished: null. The field is an optional integer; the usual and supported pattern is to leave the field out when you don’t want TTL. Sending explicit null can fail API validation (wrong type for a non-nullable int), depending on version and strictness.

robholland:
@PurseChicken do you have an example of this not working, like a validation tool/site that would reject this? The json schema for the field is:

```json
      "format": "int32",
      "type": [
        "integer",
        "null"
      ]
```

PurseChicken:
@robholland Thanks! 
You're right — the schema clearly allows null there (`["integer", "null"]`). I've validated the chart works with it, so this isn't a real concern. My mistake.
I was generalizing from stricter validators I've hit on other fields where the schema wasn't nullable, but that doesn't apply here. Thanks again!

PR #879, opened 2026-04-06
Add ttlSecondsAfterFinished to the schema job and allow setting to nu…

GBlodgett35:
…ll to disable

## What was changed

Allow for configuring the ttlSecondsAfterFinished for the server job. You can set it to null to disable it entirely. 

## Why?

For users who use ArgoCD, this TTL causes applications to go out of sync. This causes alerts to fire and masks any other issues in the environment. Allowing users to disable this via the helm chart would allow users to more effectively use this helm chart with the gitops approach. 

## Checklist

2. How was this tested:

Tested with various settings of ttlSecondsAfterFinished:

```
# Values.yaml:
ttlSecondsAfterFinished: null

# Server-job.yaml:
# ttlSecondsAfterFinished missing
```

And

```
# Values.yaml:
ttlSecondsAfterFinished: 5

# Server-job.yaml:
spec:
  backoffLimit: 100
  ttlSecondsAfterFinished: 5
```

3. Any docs updates needed?
NA - the README does not contain anything about the schema job right now

### DIFF
--- a/charts/temporal/templates/server-job.yaml
+++ b/charts/temporal/templates/server-job.yaml
@@ -7,7 +7,9 @@ metadata:
     {{- include "temporal.resourceLabels" (list $ "database" "") | nindent 4 }}
 spec:
   backoffLimit: {{ $.Values.schema.setup.backoffLimit }}
-  ttlSecondsAfterFinished: 86400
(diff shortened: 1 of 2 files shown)


robholland:
v1 is now released, we won't be back porting fixes.

PR #885, opened 2026-04-10
Add a note to README about how externalSecret should be configured.

### LINKED ISSUE #799: Please document how to use an existingSecret for DB connection parameters
There is no documentation on how to use an existing K8s secret.  I would like some example like is given in the OpenFGA helm chart under the heading

```Using an existing secret for Postgres or MySQL```

https://artifacthub.io/packages/helm/openfga/openfga

Passing the plaintext password directly to the helm chart should be considered insecure and not used in a production deployment.

### PR DESCRIPTION
**Pull request by robholland**
## What was changed
Documented an example of using externalSecret.

## Checklist

1. Closes #799

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,57 @@ server:
 
 See the example values files in the `values/` directory for complete examples.
 
+#### Using an existing Kubernetes secret
+
+If you have already created a Kubernetes secret containing your database password, reference it with `existingSecret` and `secretKey`:
+
+```bash
+# Create the secret before installing the chart
+kubectl create secret generic temporal-db-secret \
+  --from-literal=password=your_db_password
+```
+
+The secret should have the following structure:
+
+```yaml
+apiVersion: v1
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -105,6 +105,57 @@ server:
...
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: temporal-db-secret
+type: Opaque
+data:
+  password: <base64-encoded-password>
+```
```
jiechenz: nit: This YAML block could be confusing as something the user needs to write. If we want to keep it, how about rephrase to "To verify the secret was created, run kubectl get secret temporal-db-secret -o yaml" so readers understand this is output, not a second step.

jiechenz (approved):

README.md
robholland: I've re-worded, please see if it's better now.

jiechenz (approved):

PR #901, opened 2026-04-27
Omit Elasticsearch cluster settings

### LINKED ISSUE #896: [Feature Request] Support skipping cluster settings when managing Elasticsearch schema
## Problem

When `manageSchema: true` is set for the Elasticsearch visibility store, the schema job applies `cluster_settings_v7.json` which sets:

```json
{
  "persistent": {
    "action.auto_create_index": "false"
  }
}
```

This disables auto-creation of indices **cluster-wide**, which is problematic when Elasticsearch is shared with other services that rely on automatic index creation.

Currently the only workaround is to set `manageSchema: false` and manually create the index template and visibility index, losing the benefit of automatic schema management on upgrades.

## Proposal

Add a configuration option (e.g., `skipClusterSettings: true`) to allow users to skip applying `cluster_settings_v7.json` while still managing the index template and index creation.

Example usage:

```yaml
server:
  config:
    persistence:
      datastores:
        visibility:
          elasticsearch:
            version: v7
            url:
              scheme: http
              host: "shared-elasticsearch:9200"
            indices:
              visibility: temporal_visibility_v1
            manageSchema: true
            skipClusterSettings: true  # new option
```

This would allow the schema job to:
- ✅ Create/update the index template
- ✅ Create the visibility index
- ✅ Update index mappings on upgrade
- ❌ Skip applying cluster-wide settings

## Use Case

Shared Elasticsearch clusters where Temporal is one of many tenants. Disabling `auto_create_index` globally breaks other services.

stuart-wells:
https://github.com/temporalio/helm-charts/issues/896

Omitting Elasticsearch cluster settings 

## What was changed
`auto_create_index: false` can cause issues for shared Elasticsearch clusters. This is the only setting in https://github.com/temporalio/temporal/blob/main/schema/elasticsearch/visibility/cluster_settings_v7.json , so applying these settings is now being omitted.

## Why?
From the issue "This disables auto-creation of indices cluster-wide, which is problematic when Elasticsearch is shared with other services that rely on automatic index creation."

## Checklist

1. Closes #896

2. How was this tested:
- Checked helm template results
- Ran local kind cluster and verified elasticsearch auto_create_index was no longer overwritten.

3. Any docs updates needed?
No


charts/temporal/templates/server-job.yaml
```diff
@@ -97,7 +97,11 @@ spec:
                 {{- fail (printf "manageSchema is not supported for %s SQL database plugin" $store.config.pluginName) -}}
               {{- end }}
             {{- else if eq $store.driver "elasticsearch" }}
+              {{- if $store.config.skipClusterSettings }}
+            - temporal-elasticsearch-tool create-index && temporal-elasticsearch-tool update-schema --index "$ES_VISIBILITY_INDEX"
```
robholland: This isn't quite right, if we `create-index` without the template being there then the index will lack the proper settings/mappings.

robholland (changes requested):
Having looked at the cluster settings (https://github.com/temporalio/temporal/blob/main/schema/elasticsearch/visibility/cluster_settings_v7.json) it's only the one setting, and we don't need that in helm as we enforce ordering via hooks so that the index is created before it gets used by the service. Let's just omit the cluster settings unconditionally. I think `elasticsearch-tool update-schema` without `--index` will create the template if it doesn't exist.

charts/temporal/tests/server_job_test.yaml
```diff
@@ -585,3 +585,53 @@ tests:
       - equal:
           path: spec.template.spec.initContainers[?(@.name=="manage-schema-default-store")].env[?(@.name=="SQL_PORT")].value
           value: "5432"
+  - it: does not use setup-schema for elasticsearch
```
robholland: I don't think this is the right layer to test this, you can drop the tests.

robholland (approved):
Please update the PR description before merging.

PR #907, opened 2026-05-05
feat: skip password Secret/env when passwordCommand is set

dani29:
## Summary

Server v1.31 added [`passwordCommand`](https://github.com/temporalio/temporal/pull/9879) for SQL datastores — a shell hook re-run on every new connection, designed for short-lived credentials like AWS RDS IAM auth tokens or GCP Cloud SQL IAM tokens.

The chart already passes `passwordCommand` through `temporal.persistence.filterConfig` (it was never in the `omit` list), so it surfaces correctly in the rendered `config_template.yaml`. But operators still get an auto-generated empty `temporal-{name}-store` Secret plus a `TEMPORAL_*_STORE_PASSWORD` env on every server pod and `SQL_PASSWORD` / `CASSANDRA_PASSWORD` / `ES_PWD` on the schema-setup admin-tools Job — plumbing that's misleading when credentials are sourced from a command instead of a static secret.

This PR skips that plumbing whenever a store config has `passwordCommand` set:

- `templates/server-secret.yaml` — skip auto-Secret creation
- `templates/server-deployment.yaml` — skip `TEMPORAL_*_STORE_PASSWORD` env wiring
- `templates/_admintools-env.yaml` — skip `SQL_PASSWORD` / `CASSANDRA_PASSWORD` / `ES_PWD`
- `values.yaml` — documented the option with an AWS RDS IAM example

## Backward compatibility

When neither `passwordCommand` is set, behavior is unchanged. Verified by diffing \`helm template\` output between \`main\` and this branch on a representative values.yaml using \`existingSecret\` — empty diff (bit-for-bit identical).

## Verification

Rendered with a SQL store using \`passwordCommand\`:
- ✅ ConfigMap retains the \`passwordCommand:\` block
- ✅ No \`temporal-{name}-store\` Secret created
- ✅ No \`TEMPORAL_*_STORE_PASSWORD\` env on server Deployments
- ✅ No \`SQL_PASSWORD\` env on the schema-setup Job

Rendered with the existing \`existingSecret\` flow:
- ✅ Bit-for-bit identical to \`main\`

## Note for users

The configured \`command\` binary must be present in both the server image and (if \`schema.setup.enabled\`) the admin-tools image. The Temporal admin-tools \`temporal-sql-tool\` does not yet honor \`passwordCommand\` directly (server-only feature today) — pair this with an entrypoint wrapper or a sidecar that materializes the credential into the \`SQL_PASSWORD\` env for the schema Job.

## Test plan

- [x] \`helm template\` with \`passwordCommand\` set → no Secret, no \`*_PASSWORD\` env, \`passwordCommand\` block in ConfigMap
- [x] \`helm template\` with \`existingSecret\` set → identical to upstream \`main\`
- [ ] Field test: end-to-end deploy against AWS RDS with IAM auth (out of scope for this PR; we'll verify in our staging cluster and post results)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/charts/temporal/templates/_admintools-env.yaml
+++ b/charts/temporal/templates/_admintools-env.yaml
@@ -35,8 +35,10 @@
   value: {{ required (printf "Please specify databaseName for %s store" $store.name) $store.config.databaseName }}
 - name: SQL_USER
   value: {{ $store.config.user | quote }}
+{{- if not $store.config.passwordCommand }}
 - name: SQL_PASSWORD
   {{- include "temporal.password-env" (list $root $store) | nindent 2 }}
+{{- end }}
   {{- with $store.config.connectAttributes }}
 - name: SQL_CONNECT_ATTRIBUTES
   value: {{ include "temporal.persistence.sql.connectAttributes" . | quote }}
--- a/README.md
+++ b/README.md
@@ -97,10 +97,11 @@ server:
   - `manageSchema`: If `true`, the chart will run schema setup/upgrade jobs (default: `true`)
   - `existingSecret`: Reference to an existing Kubernetes secret containing credentials (e.g., `temporal-db-secret`). If not set, the chart will create a new secret.
   - `secretKey`: Key name within the secret to read the password from (default: `password`)
-- **Password handling**: Passwords are always stored in Kubernetes secrets and read from environment variables—they are never written to ConfigMaps or other manifests, even if you supply a plaintext `password` in values for bootstrap only.
+- **Password handling**: With `password` or `existingSecret`, passwords are stored in Kubernetes secrets and read from environment variables—they are never written to ConfigMaps or other manifests, even if you supply a plaintext `password` in values for bootstrap only.
   - If `existingSecret` is set, the chart uses that secret and ignores any `password` field in values for that datastore
   - If `existingSecret` is not set, the chart creates a secret from the `password` value in values
-  - The server configuration always reads passwords from environment variables that reference these secrets
+  - The server configuration reads passwords from environment variables that reference these secrets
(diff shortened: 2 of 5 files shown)


charts/temporal/templates/_admintools-env.yaml
```diff
@@ -10,8 +10,10 @@
   value: {{ required (printf "Please specify keyspace for %s store" $store.name) $store.config.keyspace }}
 - name: CASSANDRA_USER
   value: {{ $store.config.user }}
+{{- if not $store.config.passwordCommand }}
```
jiechenz: From https://github.com/temporalio/temporal/pull/9879/changes, `PasswordCommand` supports only for SQL but not ES or Cassandra. Pls limit the scope of suppression to SQL only, better to fail fast if `passwordCommand` appears on a non-SQL store.

jiechenz (changes requested):
Could you also help to update README?
Since it currently says passwords are always stored in Kubernetes Secrets and read from env vars, which is no longer true for `passwordCommand`.
Can we also document that passwordCommand is SQL-only, requires Temporal server v1.31+. Thanks

dani29:
Good call, thanks. Pushed in 8b7f9a8 (squashed onto the original feat commit).

Two things in there:

1. README + `values.yaml` reworded so the "always in Secrets / from env vars" line now only covers `password` and `existingSecret`, with a sub-bullet for `passwordCommand` calling out the **SQL-only** + **server v1.31+** constraints and the IAM-token use case.
2. While doing the docs I noticed the original templates were a little too generous — they'd skip the Secret/env wiring on Cassandra and Elasticsearch stores too if you happened to set `passwordCommand` there, even though upstream Temporal only consumes the field for SQL. Tightened the conditionals to gate on `$store.driver == "sql"` so non-SQL stores render unchanged. Verified with `helm template` against the `values.cassandra.yaml`, `values.elasticsearch.yaml`, and `values.mysql.yaml` fixtures — bit-for-bit parity vs `origin/main` when no `passwordCommand` is set.

Let me know if you'd prefer this as a separate fixup commit instead of squashed in.

jiechenz (approved):
lgtm, thank you!

PR #910, opened 2026-05-11
Readiness auto-switch from grpc to tcpsocket probe when using TLS

### LINKED ISSUE #894: [Bug] frontend GRPC readinessProbe cannot support TLS

Enable TLS on the temporal-frontend service.

### Describe the bug

After configuring certificates for the frontend service, the GRPC readinessProbe fails because grpc probes [do not yet](https://github.com/kubernetes/enhancements/issues/4939) support TLS.
```yaml
server:
  frontend:
    readinessProbe:
      grpc:
        port: 7233
        service: temporal.api.workflowservice.v1.WorkflowService
```
Changing to a readinessProbe using tcpSocket works fine, but the existing grpc section cannot be removed and deployment fails if trying to override with
```yaml
server:
  frontend:
    readinessProbe:
      tcpSocket:
        port: rpc
```
Suggestion is either to specify the readinessProbe only as a comment/recommendation, or alternatively checking for TLS configuration of the frontend service in the templating and switching the probe accordingly.

### Minimal Reproduction

Deploy with TLS configured for frontend service, observe frontend service does not become ready and other services are in crashloop since no frontend Pods are ready for the service.

### Environment/Versions

- OS and processor: [e.g. M1 Mac, x86 Windows, Linux] N/A
- Temporal Version: Helm Chart 1.1.1
- Are you using Docker or Kubernetes or building Temporal from source? Kubernetes

### Additional context

Also tried setting grpc to null but this reportedly only works with some Helm versions when the temporal chart is a subchart.
```yaml
server:
  frontend:
    readinessProbe:
      grpc: null
      tcpSocket:
        port: rpc
```

Thanks very much for providing this Helm Chart, your work is appreciated!

stuart-wells:
## What was changed
When TLS is enabled, switch frontend/internode to use tcpSocket readiness probes instead of grpc probes

## Why?
grpc readiness probes do not support TLS credentials: kubernetes/enhancements#4939

## Checklist

1. Closes #894 

2. How was this tested:
- Ran tests
- Added tests
- Sample kind cluster run

3. Any docs updates needed?
N/A

### DIFF
--- a/charts/temporal/templates/server-deployment.yaml
+++ b/charts/temporal/templates/server-deployment.yaml
@@ -120,9 +120,27 @@ spec:
             tcpSocket:
               port: rpc
           {{- end }}
+          {{- /* gRPC probes don't support TLS; they are switched to tcpSocket when TLS is configured. */ -}}
+          {{- $tlsKey := "" }}
+          {{- if eq $service "frontend" }}
+          {{- $tlsKey = "frontend" }}
+          {{- end }}
+          {{- if eq $service "internal-frontend" }}
+          {{- $tlsKey = "internode" }}
+          {{- end }}
+          {{- $tlsCertFile := "" }}
+          {{- if $tlsKey }}
+          {{- $tlsCertFile = dig "tls" $tlsKey "server" "certFile" "" $.Values.server.config }}
+          {{- end }}
+          {{- if $tlsCertFile }}
+          readinessProbe:
+            tcpSocket:
+              port: rpc
+          {{- else }}
           {{- with default $.Values.server.readinessProbe $serviceValues.readinessProbe }}
           readinessProbe:
           {{- toYaml . | nindent 12 }}
+          {{- end }}
           {{- end }}
           volumeMounts:
             - name: config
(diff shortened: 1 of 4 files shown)


robholland (changes requested):
I think we should default to rendering the always safe tcpProbe version in the template if nothing was set in values.yaml. If there is something in values.yaml we should always use that. Defaulting at the template layer means we don't have to worry about the inability to remove a gprc/tcp stanza during merge. We can add the grpcProbe version commented out in the values.yaml file and mention it in the README for those that aren't using TLS to get the slightly nicer behaviour.

charts/temporal/templates/server-deployment.yaml
```diff
@@ -124,9 +120,15 @@ spec:
             tcpSocket:
               port: rpc
           {{- end }}
-          {{- with default $.Values.server.readinessProbe $serviceValues.readinessProbe }}
+          {{- /* Default to tcpSocket (works with TLS); users may override per-service via server.<svc>.readinessProbe. */ -}}
+          {{- $resolvedProbe := default $.Values.server.readinessProbe $serviceValues.readinessProbe }}
+          {{- if $resolvedProbe }}
+          readinessProbe:
+          {{- toYaml $resolvedProbe | nindent 12 }}
+          {{- else if or (eq $service "frontend") (eq $service "internal-frontend") }}
```
robholland: Think this can be `ne $service "worker"` as per the `livenessProbe`.

robholland (approved):
Looks good, one tweak.

PR #912, opened 2026-05-14
add HorizontalPodAutoscaler support for server services

### LINKED ISSUE #911: [Feature Request] Temporal services need autoscaling feature
---
  Is your feature request related to a problem? Please describe.

  Temporal services should be able to scale up and down as per the HPA
  configuration.

  Currently the Temporal Helm chart only supports static replica counts via
  replicaCount per service. There is no way to configure a
  HorizontalPodAutoscaler for frontend, history, matching, or worker services
  without manually creating HPA resources outside the chart. This makes it
  difficult to handle variable workloads without manual intervention.

  Describe the solution you'd like

  Add native HorizontalPodAutoscaler (autoscaling/v2) support to the Helm chart
  for all Temporal server services (frontend, internal-frontend, history,
  matching, worker), configurable per service via values.yaml:

  server:
    frontend:
      autoscaling:
        enabled: true
        minReplicas: 1
        maxReplicas: 5
        targetCPUUtilizationPercentage: 80
        targetMemoryUtilizationPercentage: 70  # optional
        behavior:                              # optional
          scaleUp:
            stabilizationWindowSeconds: 60
          scaleDown:
            stabilizationWindowSeconds: 300
        customMetrics: []                      # optional extra autoscaling/v2
  metrics

  - Autoscaling should be disabled by default to avoid breaking existing
  deployments.
  - When enabled, spec.replicas should be omitted from the Deployment to prevent
   conflicts between the HPA and the Deployment controller.

  Additional context

  This is a common pattern supported by most Helm charts (e.g. ingress-nginx,
  cert-manager). As Temporal workloads are often bursty — high workflow
  throughput during business hours and low during off-peak times — HPA support
  would allow operators to run cost-efficiently without over-provisioning.

shankarkc:
## What was changed
  - Added `charts/temporal/templates/server-hpa.yaml` to render a
  `HorizontalPodAutoscaler` (autoscaling/v2) for `frontend`,
  `internal-frontend`, `history`, `matching`, and `worker` services
  - Added `autoscaling` config block to each server service in `values.yaml`
  (disabled by default — no breaking change)
  - Updated `server-deployment.yaml` to omit `spec.replicas` when autoscaling is
   enabled, preventing conflicts between the HPA and the Deployment controller
  - Updated `README.md` with an autoscaling configuration section including CPU,
   memory, behavior, and custom metrics examples

  ## Why?
  Without HPA support, scaling Temporal server services requires manual
  `replicaCount` changes. This change allows operators to enable
  CPU/memory-based autoscaling per service without any custom resource
  workarounds. As Temporal workloads are often bursty, HPA support allows
  running cost-efficiently without over-provisioning.

  1. Closes #911

  2. How was this tested:
  - `helm unittest charts/temporal` passes all 149 tests
  - New test suite `tests/server_hpa_test.yaml` covers: HPA disabled by default,
   per-service creation, minReplicas/maxReplicas, CPU and memory metrics,
  scaleTargetRef correctness, scaleUp behavior, scaleDown behavior, combined
  scaleUp+scaleDown, and skipping when the service itself is disabled
  - `tests/server_deployment_test.yaml` extended to verify: `spec.replicas` is
  omitted when autoscaling is enabled, per-service `replicaCount` overrides
  alongside a global `replicaCount`, global fallback for services without a
  per-service override, and HPA takes precedence when both `autoscaling.enabled:
   true` and `replicaCount` are set simultaneously

  3. Any docs updates needed?
  - `README.md` updated with a new "Enable Autoscaling (HPA)" section
  documenting all available values

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # Temporal Helm Chart
+
 [![FOSSA Status](https://app.fossa.com/api/projects/git%2Bgithub.com%2Ftemporalio%2Ftemporal-helm-charts.svg?type=shield)](https://app.fossa.com/projects/git%2Bgithub.com%2Ftemporalio%2Ftemporal-helm-charts?ref=badge_shield)
 
 > **For users upgrading from 0.x releases:** Please see [UPGRADING.md](UPGRADING.md) for important migration information and breaking changes.
@@ -16,8 +17,9 @@ This Helm Chart code is tested by a dedicated test pipeline. It is also used ext
 ## Prerequisites
 
 This sequence assumes
-* that your system is configured to access a kubernetes cluster (e. g. [AWS EKS](https://aws.amazon.com/eks/), [kind](https://kind.sigs.k8s.io/), or [minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/))
-* that your machine has the following installed and able to access your cluster:
+
+- that your system is configured to access a kubernetes cluster (e. g. [AWS EKS](https://aws.amazon.com/eks/), [kind](https://kind.sigs.k8s.io/), or [minikube](https://kubernetes.io/docs/tasks/tools/install-minikube/))
+- that your machine has the following installed and able to access your cluster:
   - [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
   - [Helm v3](https://helm.sh)
 
@@ -42,6 +44,7 @@ The sections that follow describe various deployment configurations using persis
 ### Persistence Configuration
 
 Temporal requires persistence stores for:
+
 - **Default store**: Stores workflow execution data (history, tasks, etc.)
 - **Visibility store**: Stores workflow visibility/search data
 
@@ -61,7 +64,7 @@ server:
           sql:
             createDatabase: true
             manageSchema: true
-            pluginName: mysql8  # or postgres12, postgres12_pgx
+            pluginName: mysql8 # or postgres12, postgres12_pgx
             driverName: mysql8
             databaseName: temporal
             connectAddr: "mysql.example.com:3306"
@@ -91,6 +94,7 @@ server:
 ```
 
 **Key points:**
+
 - Driver is determined by which key is present (`sql:`, `cassandra:`, or `elasticsearch:`)
 - **Helm-specific fields** (stripped before rendering to server config):
   - `createDatabase`: If `true`, the chart will create the database/keyspace if it doesn't exist (default: `true`)
@@ -191,7 +195,6 @@ extraObjects:
           key: prod/temporal/db
           property: password
 ```
-
 
 #### Example with SealedSecrets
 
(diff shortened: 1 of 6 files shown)


shankarkc:
I have read the CLA Document and I hereby sign the CLA

shankarkc:
Hello can someone look into this? This will help us a lot if we get HPA

charts/temporal/templates/server-deployment.yaml
```diff
@@ -14,7 +14,9 @@ metadata:
   labels:
     {{- include "temporal.resourceLabels" (list $ $service "deployment") | nindent 4 }}
 spec:
+  {{- if not (dig "autoscaling" "enabled" false $serviceValues) }}
```
robholland, line 17: You can just use `if $serviceValues.autoscaling.enabled`, the default values define the stanza.

charts/temporal/templates/server-hpa.yaml
```diff
@@ -0,0 +1,47 @@
+{{- if $.Values.server.enabled }}
+{{- range $service := (list "frontend" "internal-frontend" "history" "matching" "worker") }}
+{{- $serviceValues := index $.Values.server $service }}
+{{- if and $serviceValues.enabled (dig "autoscaling" "enabled" false $serviceValues) }}
```
robholland, line 4: Same here.

charts/temporal/tests/server_deployment_test.yaml
```diff
@@ -416,3 +416,116 @@ tests:
...
+        frontend:
+          replicaCount: 3
+          autoscaling:
+            enabled: false
+    asserts:
+      - equal:
+          path: spec.replicas
+          value: 3
+
+  - it: applies per-service replicaCount overrides alongside global replicaCount
```
robholland, line 450: You are setting each override, so none will fall back to global. If that is the intent I wouldn't set global or mention it in the description here.

README.md
```diff
@@ -425,6 +429,52 @@ TEMPORAL_AUTH_CLIENT_SECRET: xxxxxxxxxxxxxxx
 
 Reference: <https://docs.temporal.io/references/web-ui-server-env-vars>
 
+### Enable Autoscaling (HPA)
+
+By default all server services run with a static replica count. You can enable a `HorizontalPodAutoscaler` (autoscaling/v2) per service to let Kubernetes scale replicas automatically based on CPU or memory utilisation.
+
+Autoscaling is supported for `frontend`, `internal-frontend`, `history`, `matching`, and `worker`. When enabled, `spec.replicas` is omitted from the Deployment so the HPA is the sole replica controller.
+
```
robholland, line 437: We'd recommend scaleDown.stabilizationWindowSeconds especially for the history service which must re-balance shards when pods are added or removed. Re-balancing can take a couple of minutes, during which time there can be increased CPU across the pods due to balancing.

robholland (changes requested):
Looking good, thank you. Couple of tweaks please.

PR #915, opened 2026-05-15
feat(web): add web.temporalAddress to override TEMPORAL_ADDRESS env var

jmooo:
Hello! I just ran into this issue while trying to find a workaround to Dex/Google OAuth/temporal-ui not playing nicely when we have a custom JWT issuer for temporal to enable per-tenant namespace access to temporal resources.

It's my first contrib and I think I followed everything in CONTRIBUTING.md but let me know if I need changes, thanks!

## Summary

The `temporal-web` deployment hard-codes `TEMPORAL_ADDRESS` to point at the Temporal frontend service. There is no way to override this, which makes it impossible to interpose a proxy sidecar between the UI and the frontend without forking the chart.

## Change

Add an optional `web.temporalAddress` value. When set, it is used directly as `TEMPORAL_ADDRESS`. When unset, existing default behavior is preserved exactly, no breaking change.

## Testing

Two helm-unittest tests added to `tests/web_service_test.yaml`:
- Default produces the existing `<fullname>-frontend.<namespace>.svc:<port>` value
- Override via `web.temporalAddress` produces the custom value

All 134 tests pass (`helm unittest charts/temporal`).

## Motivation

Needed for auth proxy patterns where a sidecar intercepts the UI's outbound gRPC connection to swap authentication tokens. Without this, `additionalEnv` produces a duplicate `TEMPORAL_ADDRESS` key which ArgoCD rejects during structured merge diff validation.

### DIFF
--- a/charts/temporal/values.yaml
+++ b/charts/temporal/values.yaml
@@ -457,6 +457,7 @@ admintools:
   additionalEnv: []
   additionalEnvSecretName: ""
   additionalEnvConfigMapName: ""
+  # temporalAddress: "localhost:7233"
   # for sidecar containers, add containers here with restartPolicy: Always
   additionalInitContainers: []
   resources: {}
--- a/charts/temporal/values.yaml
+++ b/charts/temporal/values.yaml
@@ -486,7 +486,7 @@ web:
   # Override the address the UI uses to connect to the Temporal frontend.
   # Defaults to <fullname>-frontend.<namespace>.svc:<server.frontend.service.port>.
   # Useful when running a proxy sidecar that intercepts gRPC traffic.
-  # temporalAddress: ""
+  # temporalAddress: "localhost:7233"
   readinessProbe:
     initialDelaySeconds: 10
     httpGet:
--- a/charts/temporal/values.yaml
+++ b/charts/temporal/values.yaml
@@ -483,6 +483,10 @@ web:
     appProtocol: http
     annotations: {}
     # loadBalancerIP:
+  # Override the address the UI uses to connect to the Temporal frontend.
+  # Defaults to <fullname>-frontend.<namespace>.svc:<server.frontend.service.port>.
+  # Useful when running a proxy sidecar that intercepts gRPC traffic.
+  # temporalAddress: ""
   readinessProbe:
     initialDelaySeconds: 10
     httpGet:
--- a/charts/temporal/templates/_helpers.tpl
+++ b/charts/temporal/templates/_helpers.tpl
@@ -61,6 +61,36 @@ and we want to make sure that the component is included in the name.
 {{- end -}}
 
 {{/*
+Frontend service address as a short in-cluster name (host:port).
+*/}}
+{{- define "temporal.frontendAddress" -}}
+{{- printf "%s-frontend:%d" (include "temporal.fullname" .) (.Values.server.frontend.service.port | int) -}}
+{{- end -}}
+
+{{/*
+Frontend service address as a fully-qualified cluster DNS name (host.namespace.svc:port).
+*/}}
+{{- define "temporal.frontendAddress.fqdn" -}}
+{{- printf "%s-frontend.%s.svc:%d" (include "temporal.fullname" .) .Release.Namespace (.Values.server.frontend.service.port | int) -}}
+{{- end -}}
+
+{{/*
+Internal-frontend service address as a short in-cluster name (host:port).
+*/}}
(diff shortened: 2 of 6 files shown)


jmooo:
@jiechenz sorry to ping you, but you've been helping us on the Temporal discourse with our XDC setup. Do you think you could point us at the right person to review these kinds of PRs? Thanks!

charts/temporal/values.yaml
```diff
@@ -483,6 +483,10 @@ web:
     appProtocol: http
     annotations: {}
     # loadBalancerIP:
+  # Override the address the UI uses to connect to the Temporal frontend.
+  # Defaults to <fullname>-frontend.<namespace>.svc:<server.frontend.service.port>.
+  # Useful when running a proxy sidecar that intercepts gRPC traffic.
+  # temporalAddress: ""
```
jiechenz: I think an empty string wouldn't be a valid override. A non-empty address like `"localhost:7233"` could be a better example.
Could you Also add a unit test to test the empty-string edge case?

jiechenz:
> @jiechenz sorry to ping you, but you've been helping us on the Temporal discourse with our XDC setup. Do you think you could point us at the right person to review these kinds of PRs? Thanks!

sorry I just saw your message. I left a comment and will also route it to the team to review it. Thanks

charts/temporal/values.yaml
jmooo: addressed both issues, thanks for the feedback!

yijiewu-mg:
Just want to chime in that we found out the admin-tools is also hard code the `TEMPORAL_ADDRESS `, could we update that in this PR?

jmooo:
> Just want to chime in that we found out the admin-tools is also hard code the `TEMPORAL_ADDRESS `, could we update that in this PR?

Nice catch, added `admintools.temporalAddress` that works the same way as `web.temporalAddress`: set it to override, leave it unset and it falls back to the frontend service (or internal-frontend if that's enabled). Three test cases cover the override, unset fallback, and empty string fallback.

This allows an override but preserves the existing admintools logic, so should _not_ be a breaking change

robholland (changes requested):
Looking good, but getting a bit hard to read (it was on the verge before your changes). Can you add some helpers for the default and internal frontend addresses that we call from the templates so that the string building is hidden away? I think choosing which to use should stay in the templates as you have it though.

jmooo:
> Looking good, but getting a bit hard to read (it was on the verge before your changes). Can you add some helpers for the default and internal frontend addresses that we call from the templates so that the string building is hidden away? I think choosing which to use should stay in the templates as you have it though.

Nice the helpers make it look a lot better:

Extracted address building into helpers in `_helpers.tpl`: short form (`name:port`) for admintools to preserve its existing default, fqdn form (`name.namespace.svc:port`) for web which already used the longer format. Doing it this way should make the whole PR backward compatible.

Four helpers total; short and fqdn variants for both frontend and internal-frontend. Templates keep the `if/else` logic for choosing between them.

Tested on our internal infra (which has internal-frontend enabled), `TEMPORAL_ADDRESS` resolves to `temporal-internal-frontend:7236` as expected.

Thanks for the review!

robholland (approved):

yijiewu-mg:
@jmooo Thanks! Great work here appreciate you working on this

yijiewu-mg:
I just found another place where we hardcodes the Frontend address:
https://github.com/temporalio/helm-charts/blob/main/charts/temporal/templates/server-configmap.yaml#L133

```
publicClient:
  hostPort: "{{ include "temporal.componentname" (list $ "frontend") }}:{{ $server.frontend.service.port }}"
```

This gets mounted into all four server services (frontend, history, matching, worker). For folks running a service mesh like Istio, we need this to go through a mesh hostname rather than the raw k8s service to stay within the same region. For context, the reason for that is for our current setup, we have Istio multi cluster enabled. So for address like `temporal-frontend:7233`, the traffic could get routed to another cluster in different region.


Wondering if it would make sense to override that as well following the same pattern as web.temporalAddress / admintools.temporalAddress? I would love to hear about the thoughts from Temporal folks.
@robholland

jmooo:
hey @yijiewu-mg, if that's something you'd like added I made https://github.com/temporalio/helm-charts/pull/942 following the same pattern as this PR

yijiewu-mg:
@jmooo Thanks! I already connect with temporal folks and ask for their feedback on your PR: https://github.com/temporalio/helm-charts/pull/942

PR #916, opened 2026-05-16
fix: use constant name for schema job to enable proper hook cleanup

celestegambardella:
The schema job name previously included Chart.Version and Release.Revision, which caused the job name to change on every upgrade. This prevented the 'before-hook-creation' delete policy from cleaning up old jobs, since Helm treats differently-named resources as unrelated. This PR makes the job name stable when Helm hooks are enabled so cleanup works during upgrades.

## What was changed

The schema job name is now conditional on `schema.useHelmHooks`:

- When `schema.useHelmHooks=true` (default): a **constant** name based only on the release name, so Helm can delete/replace the prior hook Job on upgrade. The hook delete policy also gains `hook-succeeded` so the Job is cleaned up after it completes.
- When `schema.useHelmHooks=false`: the **original dynamic** name (including `Chart.Version` and `Release.Revision`) is preserved, so the immutable Job is replaced on upgrade.

### Before

```yaml
{{ $jobName := include "temporal.componentname" (list $ (printf "schema-%s-%d" $.Chart.Version $.Release.Revision | replace "." "-" | replace "+" "-")) }}
```

### After

```yaml
{{ $jobName := ternary (printf "%s-schema" (include "temporal.fullname" $)) (include "temporal.componentname" (list $ (printf "schema-%s-%d" $.Chart.Version $.Release.Revision | replace "." "-" | replace "+" "-"))) $.Values.schema.useHelmHooks }}
```

And the hook delete policy:

```yaml
"helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
```

## Why?

The dynamic job name changed on every upgrade (e.g., `temporal-schema-0-62-0-1` → `temporal-schema-0-63-0-2`). Because Helm treats differently-named resources as unrelated, `before-hook-creation` never cleaned up the old Job. This resulted in:

- Old completed jobs accumulating in the cluster
- Manual job deletion required during upgrades
- Failed upgrades when jobs with conflicting names exist

A constant name (when hooks are enabled) lets `before-hook-creation` properly clean up old jobs during upgrades. Keeping the dynamic name when hooks are disabled preserves the prior behavior for the immutable-Job-replacement path.

## Checklist

1. How was this tested:
- [x] `helm unittest charts/temporal -f 'tests/server_job_test.yaml'` — covers constant name (hooks on), dynamic name (hooks off), and the new delete policy
- [x] `helm template` renders the expected job name in both modes
- [x] Local Kind cluster: install → verify job runs → upgrade → verify old job deleted, new job created

2. Any docs updates needed?
- No README updates needed (internal template change)
- May want to mention in upgrade notes that the schema job name format changed when `useHelmHooks` is enabled

### DIFF
--- a/charts/temporal/templates/server-job.yaml
+++ b/charts/temporal/templates/server-job.yaml
@@ -1,4 +1,4 @@
-{{- $jobName := printf "%s-schema" (include "temporal.fullname" $) }}
+{{- $jobName := ternary (printf "%s-schema" (include "temporal.fullname" $)) (include "temporal.componentname" (list $ (printf "schema-%s-%d" $.Chart.Version $.Release.Revision | replace "." "-" | replace "+" "-"))) $.Values.schema.useHelmHooks }}
 apiVersion: batch/v1
 kind: Job
 metadata:
@@ -10,7 +10,7 @@ metadata:
     {{- if $.Values.schema.useHelmHooks }}
     "helm.sh/hook": pre-install,pre-upgrade
     "helm.sh/hook-weight": "0"
-    "helm.sh/hook-delete-policy": before-hook-creation
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
     {{- end }}
     {{- with $.Values.schema.jobAnnotations }}
     {{- toYaml . | nindent 4 }}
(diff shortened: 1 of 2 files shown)


DwaineSaunderson:
lgtm

robholland:
Sorry for the delay. Can you adjust to the helm hooks version please? There are common use cases where we have to turn off helm hooks and those use cases also struggle if the job name doesn't change.

celestegambardella:
@robholland I have updated this based on your recommendation thanks!

robholland (approved):

PR #929, opened 2026-06-01
feat(web,frontend): add httproute

### LINKED ISSUE #928: [Feature Request] Add support for Gateway API

Some of the clusters I manage do not use an Ingress API controller and instead have an Gateway API controller. Currently, the Chart only supports the Ingress API but not Gateway API resources.

Add support for creating HTTPRoute resources the same as for the existing Ingress resources, with being disabled by default with the option to enable them via the values.yaml

ArkShocer:
## What was changed

Added support for Gateway API, allowing the chart to create an HTTPRoute resource as an alternative to the existing Ingress configuration.

## Why?

Gateway API is becoming the standard way to manage ingress traffic in Kubernetes. With the deprecation of [ingress-nginx](https://github.com/kubernetes/ingress-nginx) in March 2026, many engineers have moved to [Envoy Gateway](https://gateway.envoyproxy.io/), which only supports the Gateway API — so charts would need to support it in order to not have the engineers manually add the manifests themself.
## Checklist

1. Closes https://github.com/temporalio/helm-charts/issues/928

2. How was this tested:
Ran `helm template` locally to verify the rendered manifest.

3. Any docs updates needed?
Don't think so as I did not find equivalent doc for Ingress, but I can add it if wanted.

### DIFF
--- a/charts/temporal/templates/frontend-httproute.yaml
+++ b/charts/temporal/templates/frontend-httproute.yaml
@@ -1,0 +1,32 @@
+{{- if .Values.server.frontend.httpRoute.enabled -}}
+apiVersion: gateway.networking.k8s.io/v1
+kind: HTTPRoute
+metadata:
+  name: {{ include "temporal.componentname" (list $ "frontend") }}
+  labels:
+    {{- include "temporal.resourceLabels" (list $ "frontend" "") | nindent 4 }}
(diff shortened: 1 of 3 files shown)


ArkShocer:
Hello can someone review the PR? Would be nice to deploy the route manifest with the chart.

PR #930, opened 2026-06-03
fix: PDB templates break on multi-key specs and ignore global replicaCount

vtomasr5:
## What was changed

Fixed two rendering bugs in the temporal chart's PodDisruptionBudget templates, plus a cosmetic cleanup:

  - **`server-pdb.yaml` / `web-pdb.yaml`**: replaced `spec:` + `  {{ toYaml .podDisruptionBudget }}` with `{{- toYaml ... | nindent 2 }}`. The old form only indented the first line of the rendered spec, so any PDB config with more than one key (e.g., `minAvailable` + `unhealthyPodEvictionPolicy`) produced invalid YAML: `Error: YAML parse error ... mapping values are not allowed in this context`.
  - **`server-pdb.yaml`**: the creation guard reads only the per-service `replicaCount`, while deployments fall back to the global `server.replicaCount` (`server-deployment.yaml`). With the default config (no per-service `replicaCount`), `nil | int` → `0` and server PDBs were silently never rendered. The guard now resolves replicas with the same global fallback via a `hasKey`-guarded ternary, so an explicit `replicaCount: 0` still suppresses the PDB.
  - Moved the `---` document separator inside the conditional so services without a PDB no longer emit empty YAML documents.
  - `tests/server_pdb_test.yaml`: added regression tests for the global-fallback and multi-key cases, and moved the suite-level `replicaCount: 1` (which masked the fallback bug) into the only test that needs it.

## Why?

Anyone setting `server.<service>.podDisruptionBudget` without also explicitly setting a per-service `replicaCount` got **no PDB at all** — silently, since the template just skips rendering. And any PDB spec with more than one key broke `helm template`/`helm install` outright. Both made the `podDisruptionBudget` values effectively unusable in their most common configurations, with no signal in the first case.

## Checklist

  1. How was this tested?
     - `helm unittest .` — 13 suites, 135 tests passed, including the new regression tests.
     - `helm template .` with each of the four `ci/*.yaml` configs renders cleanly.
     - Manual renders with multi-key PDB values for both server `frontend` and `web` confirmed valid output.
     - `helm lint` passes.

  2. Any docs updates needed?
     - Nope

### DIFF
--- a/charts/temporal/templates/server-pdb.yaml
+++ b/charts/temporal/templates/server-pdb.yaml
@@ -1,7 +1,8 @@
 {{- if $.Values.server.enabled }}
 {{- range $service := (list "frontend" "internal-frontend" "history" "matching" "worker") }}
 {{- $serviceValues := index $.Values.server $service }}
-{{- if and $serviceValues.enabled (gt ($serviceValues.replicaCount | int) 0) ($serviceValues.podDisruptionBudget) }}
+{{- $replicaCount := ternary $serviceValues.replicaCount $.Values.server.replicaCount (hasKey $serviceValues "replicaCount") }}
+{{- if and $serviceValues.enabled (gt ($replicaCount | int) 0) ($serviceValues.podDisruptionBudget) }}
 apiVersion: policy/v1
 kind: PodDisruptionBudget
 metadata:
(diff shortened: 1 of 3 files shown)


stuart-wells (approved):
Looks good to me, thanks for the contribution!

PR #941, opened 2026-06-22
Render metadata.namespace for namespaced resources

ondrejlevy:
## What was changed

This PR renders `metadata.namespace` for the namespaced resources produced by the Temporal Helm chart, using the Helm release namespace through a shared `temporal.kubernetesNamespace` helper.

The change covers Deployments, Services, ConfigMaps, Secrets, Jobs, PodDisruptionBudgets, Ingresses, ServiceMonitor, ServiceAccount, and the test Pod. It also updates same-namespace Temporal service references to use short service DNS names and removes the default ServiceMonitor `namespaceSelector`, so the ServiceMonitor follows the Prometheus Operator same-namespace default.

Automated helm-unittest coverage was added to assert namespace rendering across the chart resources, including [charts/temporal/tests/namespace_rendering_test.yaml](charts/temporal/tests/namespace_rendering_test.yaml).

## Why?

This makes the chart work reliably when rendered through Kustomize `helmCharts` with a target namespace.

Starting with Kustomize `v5.8.0`, namespace propagation to Helm charts changed, and `v5.8.1` completed the namespace propagation fix. In this flow, Kustomize passes the namespace into `helm template --namespace`, but it does not patch missing `metadata.namespace` into the generated Helm output. Therefore charts that do not explicitly render `.Release.Namespace` can still produce manifests without `metadata.namespace`.

That is a problem for GitOps and Kustomize-based deployments where rendered manifests are expected to contain the target namespace explicitly. With this change, rendering the chart through Kustomize `helmCharts` produces namespaced manifests consistently.

Use case:

- A platform team deploys Temporal through Kustomize or a GitOps controller.
- The Kustomization sets `namespace: target-ns`.
- The Temporal chart is rendered through `helmCharts`.
- The generated resources should include `metadata.namespace: target-ns` so the rendered output is self-contained and namespace-aware.

## Checklist

 How was this tested:

- `helm lint charts/temporal -f charts/temporal/ci/postgres-values.yaml`
- `helm template temporal charts/temporal --namespace target-namespace -f charts/temporal/ci/postgres-values.yaml | grep -c '^  namespace: target-namespace$'`
  - Result: `20`
- Reproduced with `kustomize v5.8.1` and `helmCharts`:
  - upstream chart rendered `0` occurrences of `namespace: target-ns`
  - this branch rendered `20` occurrences of `namespace: target-ns`
- Added helm-unittest coverage for namespace rendering and related ServiceMonitor / Temporal address behavior.

Any docs updates needed?

No. This does not add a new value or change the documented install flow. It makes the rendered manifests include the namespace Helm already receives via `--namespace`.

### DIFF
--- a/charts/temporal/templates/server-namespace-job.yaml
+++ b/charts/temporal/templates/server-namespace-job.yaml
@@ -3,6 +3,7 @@
 apiVersion: batch/v1
 kind: Job
 metadata:
+  namespace: {{ include "temporal.kubernetesNamespace" $ }}
   name: {{ $jobName }}
   labels:
     {{- include "temporal.resourceLabels" (list $ "database" "") | nindent 4 }}
@@ -38,11 +39,11 @@ spec:
           env:
             - name: TEMPORAL_ADDRESS
               {{- if (index $.Values.server "internal-frontend").enabled }}
-              value: {{ include "temporal.fullname" $ }}-internal-frontend.{{ $.Release.Namespace }}.svc:{{ (index $.Values.server "internal-frontend").service.port }}
+              value: {{ include "temporal.fullname" $ }}-internal-frontend:{{ (index $.Values.server "internal-frontend").service.port }}
               {{- else if $.Values.server.frontend.ingress.enabled }}
               value: "{{ index $.Values.server.frontend.ingress.hosts 0 }}"
               {{- else }}
-              value: "{{ include "temporal.fullname" $ }}-frontend.{{ $.Release.Namespace }}.svc:{{ $.Values.server.frontend.service.port }}"
+              value: "{{ include "temporal.fullname" $ }}-frontend:{{ $.Values.server.frontend.service.port }}"
               {{- end }}
             {{- with $.Values.admintools.additionalEnv }}
               {{- toYaml . | nindent 12 }}
--- a/charts/temporal/templates/server-namespace-job.yaml
+++ b/charts/temporal/templates/server-namespace-job.yaml
@@ -39,11 +39,11 @@ spec:
           env:
             - name: TEMPORAL_ADDRESS
               {{- if (index $.Values.server "internal-frontend").enabled }}
-              value: {{ include "temporal.fullname" $ }}-internal-frontend:{{ (index $.Values.server "internal-frontend").service.port }}
+              value: {{ include "temporal.internalFrontendAddress" $ }}
               {{- else if $.Values.server.frontend.ingress.enabled }}
               value: "{{ index $.Values.server.frontend.ingress.hosts 0 }}"
               {{- else }}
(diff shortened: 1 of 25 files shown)


charts/temporal/templates/web-deployment.yaml
```diff
@@ -39,7 +40,7 @@ spec:
           imagePullPolicy: {{ .Values.web.image.pullPolicy }}
           env:
             - name: TEMPORAL_ADDRESS
-              value: "{{ include "temporal.fullname" $ }}-frontend.{{ .Release.Namespace }}.svc:{{ .Values.server.frontend.service.port }}"
```
stuart-wells: Looks like there's a merge conflict in this section. I think this could be fixed by taking main and shortening temporal.frontendAddress.fqdn to temporal.frontendAddress in the `else`

charts/temporal/templates/server-namespace-job.yaml
```diff
@@ -38,11 +39,11 @@ spec:
           env:
             - name: TEMPORAL_ADDRESS
               {{- if (index $.Values.server "internal-frontend").enabled }}
-              value: {{ include "temporal.fullname" $ }}-internal-frontend.{{ $.Release.Namespace }}.svc:{{ (index $.Values.server "internal-frontend").service.port }}
```
stuart-wells, line 41: nit: There are `temporal.frontendAddress` and `temporal.internalFrontendAddress` helpers that can shorten this and the below change

charts/temporal/templates/shim-configmap.yaml
```diff
@@ -2,15 +2,11 @@
...
-  name: "{{ include "temporal.fullname" $ }}-shims"
-  labels:
-    {{- include "temporal.resourceLabels" (list $ "" "") | nindent 4 }}
-  {{- if $.Values.schema.useHelmHooks }}
-  annotations:
-    "helm.sh/hook": pre-install,pre-upgrade
-    "helm.sh/hook-weight": "-1"
-    "helm.sh/hook-delete-policy": before-hook-creation
-  {{- end }}
+{{- printf "namespace: %s\nname: \"%s-shims\"\nlabels:" (include "temporal.kubernetesNamespace" $) (include "temporal.fullname" $) | nindent 2 }}
```
stuart-wells: nit:  This can be simplified to `namespace: {{ include "temporal.kubernetesNamespace" $ }}` and `name: "{{ include "temporal.fullname" $ }}-shims"` can be left alone, to match the other files in this PR

stuart-wells (changes requested):
Thanks for the contribution! Overall this looks good, I'd just ask that you handle the merge conflict and a couple nits before approval.

stuart-wells:
Completed in https://github.com/temporalio/helm-charts/pull/972

PR #942, opened 2026-06-29
feat(server): add server.publicClientHostPort to override publicClient.hostPort

jmooo:
## What changed?

Added `server.publicClientHostPort` — an optional value to override the `publicClient.hostPort` field in the server configmap, which is mounted into all four server services (frontend, history, matching, worker).

## Why?

The `publicClient.hostPort` is currently hardcoded to the raw Kubernetes service name (`temporal-frontend:7233`). In service mesh setups like Istio multi-cluster, raw k8s service names can route traffic across regions unexpectedly. Users need to specify a mesh-aware hostname to ensure traffic stays within the same region.

Requested in #915 (comment) by @yijiewu-mg.

## How?

Follows the same pattern as `web.temporalAddress` and `admintools.temporalAddress` from #915:
- When `server.publicClientHostPort` is set, it is used directly
- When unset, existing behavior is preserved: defaults to the frontend service address when `internal-frontend` is disabled, omitted when `internal-frontend` is enabled
- The override works regardless of whether `internal-frontend` is enabled

## How did you test it?

- [x] Three new test cases in `tests/server_configmap_test.yaml`:
  - Override is used when `server.publicClientHostPort` is set
  - Falls back to default when unset
  - Override works even when `internal-frontend` is enabled

### DIFF
--- a/charts/temporal/templates/server-configmap.yaml
+++ b/charts/temporal/templates/server-configmap.yaml
@@ -128,7 +128,10 @@ data:
       {{- toYaml . | nindent 6 }}
     {{- end }}
 
-    {{- if not (index $.Values.server "internal-frontend").enabled }}
+    {{- if $.Values.server.publicClientHostPort }}
+    publicClient:
+      hostPort: "{{ $.Values.server.publicClientHostPort }}"
+    {{- else if not (index $.Values.server "internal-frontend").enabled }}
     publicClient:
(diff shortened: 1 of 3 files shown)


yijiewu-mg:
Awesome! you are fast :)

jiechenz (approved):
lgtm

Issue #626, opened 2025-01-16
[Feature Request] Add Support for Service-Specific Service Accounts in Temporal Helm Chart


seeker815:
### Is your feature request related to a problem? Please describe.

If we enable `serviceAccount: true` helm chart creates one service account for all deploys { temporal-admintools, temporal-frontend, temporal-history, temporal-matching, temporal-web, temporal-worker } 

We need a way to  enable service account creation and customization for each service and allow the user to set unique serviceAccount names per service.

This is an issue from security point of view and also as we use consul service mesh, it mandates using same name for service account as service name when we use ACL and setting common service account name doesn't work.

### Describe the solution you'd like

we would like individual service account to be created for each of the service like so
```
  temporal-frontend:
    serviceAccount:
      create: true
      name: temporal-frontend
      extraAnnotations:
        
  temporal-history:
    serviceAccount:
      create: true
      name: temporal-history
      extraAnnotations:
        
  matching:

aman-kumar29:
Hi @seeker815 @robholland is this issue something which i can work on? if yes i'd like to spend some time on adding this feature here. Thanks.

aman-kumar29:
Hi @robholland can you please review this PR?

kriyanshii:
@robholland Can i work on this?

robholland:
Closed by #944

PR #944, opened 2026-06-30
add service specific service account overrides if required

aman-kumar29:
## What was changed

This PR allows configuring separate Kubernetes **ServiceAccounts** for each Temporal component (frontend, history, matching, worker, admintools, web, schema, test).

---

# Why We Made These Changes

## Problem

The Temporal Helm chart only supported a **single global ServiceAccount** for all components. Every pod—including **frontend, history, matching, worker, admintools, web, schema jobs, and tests**—used the same ServiceAccount.

## Why That's a Problem

In production Kubernetes environments, different components often require different cloud permissions.

For example, on **AWS EKS**, IAM roles are associated with Kubernetes ServiceAccounts using **IAM Roles for Service Accounts (IRSA)**.

Some examples:

* **Frontend** may need access to an API Gateway.
* **Worker** may require access to S3 for archival.
* **Schema jobs** may require elevated database privileges.

Using a single ServiceAccount forces every component to receive the **union of all required permissions**, violating the **Principle of Least Privilege (PoLP)** and increasing the blast radius in case of compromise.

This limitation was also requested to be addressed in **Issue #626**.
---

# Checklist

* [x] **`_helpers.tpl`**

  * Added `componentServiceAccountName` and `componentServiceAccount` helper functions.

* [x] **`serviceaccount.yaml`**

  * Creates component-specific ServiceAccounts when configured.

* [x] **`values.yaml`**

  * Added an optional `serviceAccount: {}` block for all supported components:

    * frontend
    * internal-frontend
    * history
    * matching
    * worker
    * admintools
    * web
    * schema
    * test

* [x] **`server-deployment.yaml`**

  * Uses component-specific ServiceAccounts for server components.

* [x] **`admintools-deployment.yaml`**

  * Uses component-specific ServiceAccounts.

* [x] **`web-deployment.yaml`**

  * Uses component-specific ServiceAccounts.

* [x] **`server-job.yaml`**

  * Uses component-specific ServiceAccounts for schema jobs.

* [x] **`server-namespace-job.yaml`**

  * Uses component-specific ServiceAccounts for namespace setup jobs.

* [x] **`test.yaml`**

  * Uses component-specific ServiceAccounts for test pods.

* [x] Added **46 unit tests** across **6 test files**.

* [x] All **188 Helm unit tests** pass.

* [x] `helm template` renders successfully.

* [x] `helm lint` passes.

* [x] Fully **backward compatible** — default behavior remains unchanged.

* [x] Follows Helm chart best practices and is **backward compatible** — defaults to existing behavior.

* [x] Matches the accepted contribution category:

  * **Service Account and RBAC Customization**

* [x] Open the PR referencing **Issue #626**.

---


aman-kumar29:
Hi @robholland @jiechenz, can you perhaps look into this change?

mrdimidium:
Hi, that would be useful for me — for security reasons, I wouldn't want to grant RDS access to any pods other than the necessary ones.

aman-kumar29:
yeah, I am hoping to get a response on this so that this changes gets merged.

charts/temporal/templates/serviceaccount.yaml
```diff
@@ -12,3 +12,49 @@ metadata:
       {{- toYaml . | nindent 4 }}
     {{- end }}
 {{- end -}}
+
+{{- range $service := (list "frontend" "internal-frontend" "history" "matching" "worker") }}
+{{- $serviceValues := index $.Values.server $service }}
+{{- $componentSA := dict }}
```
robholland: Should probably check the service is enabled here. There is no point creating a serviceAccount for an unused service, there will be no deployment to use it.

charts/temporal/templates/_helpers.tpl
```diff
@@ -43,12 +43,52 @@ Create the name of the service account
...
+{{- if or (eq $component "admintools") (eq $component "web") (eq $component "schema") (eq $component "test") -}}
+  {{- if hasKey (index $global.Values $component) "serviceAccount" -}}
+    {{- $componentSA = index $global.Values $component "serviceAccount" -}}
+  {{- end -}}
+{{- else -}}
+  {{- if and (hasKey $global.Values.server $component) (hasKey (index $global.Values.server $component) "serviceAccount") -}}
+    {{- $componentSA = index $global.Values.server $component "serviceAccount" -}}
+  {{- end -}}
+{{- end -}}
+{{- if and $componentSA (kindIs "map" $componentSA) (hasKey $componentSA "create") -}}
```
robholland: No `create` key should be equivalent to `create: false`. It would be surprising to have a map with `name` set that gets ignored because the `create` key isn't specifically set to `false`.

charts/temporal/templates/serviceaccount.yaml
```diff
@@ -12,3 +12,49 @@ metadata:
...
+  annotations:
+    helm.sh/hook: pre-install, pre-upgrade
+    helm.sh/hook-weight: "-10"
+    {{- with $componentSA.extraAnnotations }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
+{{- end }}
+{{- end }}
+
+{{- range $component := (list "admintools" "web" "schema" "test") }}
```
robholland: Can we push the server service detection by name and the `map`/`create` dance into a helper, so that we don't repeat the pattern twice in `componentServiceAccountName` helper and twice here. We can then also just have one copy of the template here, looping over all the server services + `admintools` et al in.

robholland (changes requested):
All conceptually good, but please refactor a little to make it easier to maintain. Also please ensure trailing newlines are present.

charts/temporal/templates/serviceaccount.yaml
aman-kumar29: let me make the changes

aman-kumar29:
hi @robholland, addressed those issues and added the repetitive code logic to helpers.tpl

robholland:
Thanks, please also fix the trailing new lines. You can see which files lack them from looking at: https://github.com/temporalio/helm-charts/pull/944/changes

aman-kumar29:
Hi @robholland, added the trailing new lines issue. can you check now?

aman-kumar29:
Hey @robholland, reminder in case you missed this. I have rebased this branch to main

robholland:
Thanks, looks good, one merge marker left in though: https://github.com/temporalio/helm-charts/pull/944/changes#diff-5bb342dc773dd7a8d75ab679927027b7149d4c127a7d65644fed63ab311d4938R589

aman-kumar29:
done @robholland

robholland (approved):

PR #951, opened 2026-07-17
Add temporal-proxy Helm chart

pseudomuto:
Introduces a new chart that deploys the Temporal proxy (https://github.com/temporalio/temporal-proxy) as a Deployment fronted by a ClusterIP Service, with runtime configuration supplied through a ConfigMap mounted at /etc/temporal-proxy/config.yaml. The chart ships as 0.1.0-alpha1.

Wires the chart into CI. helm-unittest now runs against all charts (charts/*) instead of only temporal, and ct install excludes temporal-proxy because it needs a running Temporal frontend to become healthy; ct change detection is otherwise preserved. _**I will address this in the future to run a simple dev server in a sidecar or something.**_

Reworks publishing so charts release independently. The existing release workflow now packages only the temporal chart and passes skip_packaging to chart-releaser, which otherwise auto-discovers and publishes every chart under charts/. A new publish-proxy workflow releases the proxy on demand: bump version/appVersion in Chart.yaml via PR, then run the workflow to tag and publish whatever is committed.

Also adds a (optional) mise config pinning helm and providing render/test tasks for local chart development.


charts/temporal-proxy/templates/service.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "temporal-proxy.fullname" . }}
+  namespace: {{ include "temporal-proxy.namespace" . }}
+  labels:
+    {{- include "temporal-proxy.labels" . | nindent 4 }}
+spec:
+  type: {{ .Values.service.type }}
```
pseudomuto, line 9: Currently defaults to ClusterIP. However, we may want an additional headless service (or a way to add additional manifests) so that Temporal workers can bypass kube-proxy (L4) and use its native client-side load balancing.

Punting for now, but wanted to call it out for later.

charts/temporal-proxy/templates/deployment.yaml
```diff
@@ -0,0 +1,89 @@
...
+  template:
+    metadata:
+      {{- with .Values.podAnnotations }}
+      annotations:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      labels:
+        {{- include "temporal-proxy.labels" . | nindent 8 }}
+        {{- with .Values.podLabels }}
+        {{- toYaml . | nindent 8 }}
```
robholland: nit: would indent this.

mise.toml
```diff
@@ -0,0 +1,46 @@
...
+run = '''
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Skip if already installed
+if helm plugin list | awk '$1 == "unittest" && $2 == "{{vars.version}}" { found = 1 } END { exit !found }'; then
+  exit 0
+fi
+
+helm plugin uninstall unittest 2>/dev/null || true
```
robholland, line 43: If we skip if it's already installed then we are never going to un-install it here.

robholland (approved):
Looks good. Nothing blocking.

charts/temporal-proxy/templates/deployment.yaml
pseudomuto: Done

mise.toml
pseudomuto: If a different version is installed, the check fails, and so does the subsequent installation. While useless in the initial case, this makes upgrading later work without issue.

robholland: ah makes sense. Missed the specific version.

PR #953, opened 2026-07-21
feat(temporal-proxy): add PodDisruptionBudget and extraObjects support

### LINKED ISSUE #952: temporal-proxy chart: add PodDisruptionBudget and extraObjects support (parity with the main temporal chart)
**Is your feature request related to a problem? Please describe.**

PR #951 recently added a new subchart, `charts/temporal-proxy` (currently `version: 0.1.0-alpha1`), wrapping the new `temporalio/temporal-proxy` project — a gRPC proxy that sits between Temporal SDK clients, workers, and the Temporal UI on one side and one or more upstream Temporal clusters on the other (local dev, self-hosted, Temporal Cloud, or a mix), handling namespace translation and TLS termination.

Because it sits in front of every client/worker/UI connection to one-or-more upstream clusters, it's exactly the kind of component that needs first-class production-readiness knobs. Comparing it to the mature `charts/temporal` chart in the same repo shows a gap:

- `charts/temporal/templates/server-pdb.yaml` and `web-pdb.yaml` render a `PodDisruptionBudget` per component, gated by `server.<service>.podDisruptionBudget` / `web.podDisruptionBudget` values.
- `charts/temporal/templates/extra-objects.yaml` is a generic escape hatch:
  ```yaml
  {{- range .Values.extraObjects }}
  ---
  {{ tpl . $ }}
  {{- end }}
  ```
  documented as the sanctioned way to inject `ExternalSecret`/`SealedSecret`/`NetworkPolicy`/anything else via GitOps without forking the chart.
- `charts/temporal-proxy/templates/` (`configmap.yaml`, `deployment.yaml`, `hpa.yaml`, `service.yaml`, `serviceaccount.yaml`) has **neither** a PDB template **nor** an `extraObjects` template, and `charts/temporal-proxy/values.yaml` has no corresponding keys.

Today, anyone running `temporal-proxy` cannot (a) protect it with a PDB during node drains/rolling upgrades — even though it's a single point of failure for all Temporal traffic if under-replicated during disruption — or (b) inject an `ExternalSecret`/`SealedSecret` for the mTLS client credentials it will need to reach Temporal Cloud or another upstream, without forking the chart.

**Describe the solution you'd like**

Add two small, additive templates to `charts/temporal-proxy`, following the exact pattern already proven in `charts/temporal`:

1. **`charts/temporal-proxy/templates/pdb.yaml`** (new):
   ```yaml
   {{- if .Values.podDisruptionBudget }}
   apiVersion: policy/v1
   kind: PodDisruptionBudget
   metadata:
     name: {{ include "temporal-proxy.fullname" . }}
     namespace: {{ include "temporal-proxy.namespace" . }}
     labels:
       {{- include "temporal-proxy.labels" . | nindent 4 }}
   spec:
     {{- toYaml .Values.podDisruptionBudget | nindent 2 }}
     selector:
       matchLabels:
         {{- include "temporal-proxy.selectorLabels" . | nindent 6 }}
   {{- end }}
   ```
   with `podDisruptionBudget: {}` added to `values.yaml` (mirroring `web.podDisruptionBudget: {}` in the main chart), so it doesn't render unless configured. The `_helpers.tpl` for this chart already defines `temporal-proxy.namespace`, `temporal-proxy.fullname`, `temporal-proxy.labels`, and `temporal-proxy.selectorLabels` — the exact helpers needed, no new helper work required.

   Worth noting: the sibling PDB templates on the main chart have had several bugs over their history (#572 wrong scope var, #752/#777/#778 replicaCount-gating issues, #930 multi-key indent bug, all since fixed). I'll make sure this new template avoids those exact pitfalls from day one — gate on `.Values.podDisruptionBudget` truthiness rather than a brittle replicaCount comparison, and use `toYaml ... | nindent 2` correctly under `spec:`.

2. **`charts/temporal-proxy/templates/extra-objects.yaml`** (new) — a verbatim copy of the main chart's pattern, with `extraObjects: []` added to `values.yaml`. This unlocks GitOps injection of `ExternalSecret`/`SealedSecret` (for upstream mTLS client certs), `NetworkPolicy`, `ServiceMonitor`, or anything else operators need, without waiting for the chart to grow bespoke support for each one — the same rationale the README already gives for `extraObjects` on the main chart.

3. Add `helm-unittest` coverage for both (the repo already runs `helm unittest` in CI per #951) — PDB renders/doesn't-render correctly, and `extraObjects` renders arbitrary manifests with `tpl` templating applied.

**Describe alternatives you've considered**

- Waiting for the chart to mature further before adding these — but PDB and `extraObjects` are foundational, low-risk, purely additive patterns already proven on the main chart; there's no real benefit to deferring them, and they don't lock in any design decisions that later `temporal-proxy` chart changes would need to work around.
- A bespoke NetworkPolicy or TLS-cert-mounting template instead of the generic `extraObjects` escape hatch — bigger design surface, more opinionated, and the main chart already demonstrates that a generic escape hatch is the preferred pattern here.

**Additional context**

This chart is very new (one commit at the time of writing, PR #951, merged 2026-07-20) and still `alpha1` — I'm opening this as a scoping issue first, per the repo's contribution norms, rather than jumping straight to a PR, since maintainers may still be actively restructuring `temporal-proxy`'s templates and might prefer to add this themselves or guide the exact approach. Happy to contribute a PR for this once there's a green light on the approach.

wankhede04:
Closes #952

## What

Adds two small, additive templates to `charts/temporal-proxy`, mirroring patterns already proven on the main `charts/temporal` chart:

- `templates/pdb.yaml` — an optional `PodDisruptionBudget`, gated purely on `.Values.podDisruptionBudget` truthiness (no brittle `replicaCount` comparison), using the chart's existing `temporal-proxy.fullname`/`namespace`/`labels`/`selectorLabels` helpers.
- `templates/extra-objects.yaml` — a verbatim copy of the main chart's generic escape hatch (`{{- range .Values.extraObjects }} --- {{ tpl . $ }} {{- end }}`) for injecting arbitrary manifests (e.g. `ExternalSecret`/`SealedSecret` for upstream mTLS credentials, `NetworkPolicy`) via GitOps without forking the chart.
- Corresponding `values.yaml` keys: `podDisruptionBudget: {}` and `extraObjects: []`.
- `helm-unittest` coverage for both, following this chart's existing test conventions (`charts/temporal-proxy/tests/`).

## Why

`temporal-proxy` sits in front of every SDK client/worker/UI connection to one-or-more upstream Temporal clusters (including Temporal Cloud), making it a network chokepoint. Despite that, it currently has none of the production-readiness patterns already established on the mature `charts/temporal` chart in this repo — operators can't protect it with a PDB during node drains/rolling upgrades, nor inject GitOps-managed secrets (e.g. mTLS client credentials) without forking the chart.

## How tested

- `helm lint charts/temporal-proxy` — clean
- `helm template charts/temporal-proxy` — renders neither resource by default (no-op)
- `helm template charts/temporal-proxy --set podDisruptionBudget.minAvailable=1` — PDB renders with correct `spec`/`selector`
- `helm template charts/temporal-proxy` with an `extraObjects` entry — renders the arbitrary manifest with `tpl` templating applied
- `helm unittest charts/temporal-proxy` — all suites pass, including two new ones (`pdb_test.yaml`, `extra_objects_test.yaml`)
- `helm unittest charts/*` — full repo test suite passes (21 suites, 188 tests)
- Verified each commit is independently valid (lints and passes tests in isolation)

## Risks

Purely additive — both new templates are no-ops unless explicitly configured, so this carries no risk to existing `temporal-proxy` installations.

### DIFF
--- a/charts/temporal-proxy/tests/extra_objects_test.yaml
+++ b/charts/temporal-proxy/tests/extra_objects_test.yaml
@@ -1,0 +1,75 @@
+suite: extraObjects test
+templates:
+  - extra-objects.yaml
+chart:
+  version: 1.0.0
+  appVersion: 1.0.0
+release:
+  name: temporal-proxy
+  namespace: temporal-proxy
+tests:
+  - it: does not render anything when extraObjects is empty
+    asserts:
+      - hasDocuments:
+          count: 0
+
+  - it: renders a ConfigMap from a string entry
+    set:
+      extraObjects:
+        - |
+          apiVersion: v1
+          kind: ConfigMap
+          metadata:
+            name: test-config
+          data:
+            key: value
+    asserts:
+      - containsDocument:
+          kind: ConfigMap
+          apiVersion: v1
+      - equal:
+          path: metadata.name
+          value: test-config
+
+  - it: renders an ExternalSecret from a string entry
+    set:
+      extraObjects:
+        - |
+          apiVersion: external-secrets.io/v1beta1
+          kind: ExternalSecret
+          metadata:
+            name: test-secret
+          spec:
+            secretStoreRef:
+              name: test-store
+              kind: SecretStore
+            target:
+              name: test-target
+            data:
+            - secretKey: password
+              remoteRef:
+                key: test/key
+                property: password
+    asserts:
+      - containsDocument:
+          kind: ExternalSecret
+          apiVersion: external-secrets.io/v1beta1
+      - equal:
+          path: metadata.name
+          value: test-secret
+
+  - it: supports Go templating within entries
+    set:
+      extraObjects:
+        - |
+          apiVersion: v1
+          kind: ConfigMap
+          metadata:
+            name: {{ .Release.Name }}-config
+    asserts:
+      - containsDocument:
+          kind: ConfigMap
+          apiVersion: v1
+      - equal:
+          path: metadata.name
+          value: temporal-proxy-config
--- a/charts/temporal-proxy/tests/extra_objects_test.yaml
+++ b/charts/temporal-proxy/tests/extra_objects_test.yaml
@@ -73,3 +73,32 @@ tests:
       - equal:
           path: metadata.name
           value: temporal-proxy-config
+
+  - it: renders multiple extra objects
+    set:
+      extraObjects:
+        - |
+          apiVersion: v1
+          kind: ConfigMap
+          metadata:
+            name: test-config-1
+          data:
+            key: value1
+        - |
+          apiVersion: v1
+          kind: ConfigMap
+          metadata:
+            name: test-config-2
+          data:
+            key: value2
+    asserts:
+      - hasDocuments:
+          count: 2
+      - equal:
+          path: metadata.name
+          value: test-config-1
+        documentIndex: 0
+      - equal:
+          path: metadata.name
+          value: test-config-2
+        documentIndex: 1
--- a/charts/temporal-proxy/values.yaml
+++ b/charts/temporal-proxy/values.yaml
@@ -127,3 +127,8 @@ volumeMounts: []
 nodeSelector: {}
 tolerations: []
 affinity: {}
+
+# This is for setting up a PodDisruptionBudget more information can be found here:
+# https://kubernetes.io/docs/concepts/workloads/pods/disruptions/
+podDisruptionBudget: {}
+  # minAvailable: 1
--- a/charts/temporal-proxy/values.yaml
+++ b/charts/temporal-proxy/values.yaml
@@ -132,3 +132,6 @@ affinity: {}
 # https://kubernetes.io/docs/concepts/workloads/pods/disruptions/
 podDisruptionBudget: {}
   # minAvailable: 1
+
+# Array of extra K8s manifests to deploy, e.g. an ExternalSecret for upstream mTLS credentials.
+extraObjects: []
--- a/charts/temporal-proxy/values.yaml
+++ b/charts/temporal-proxy/values.yaml
@@ -134,4 +134,6 @@ podDisruptionBudget: {}
   # minAvailable: 1
 
 # Array of extra K8s manifests to deploy, e.g. an ExternalSecret for upstream mTLS credentials.
+# Note: unlike the chart's other templates, these objects are rendered as-is and do not
+# automatically inherit the release namespace, labels, or other chart-wide values.
 extraObjects: []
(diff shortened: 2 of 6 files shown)


charts/temporal-proxy/values.yaml
```diff
@@ -127,3 +127,11 @@ volumeMounts: []
 nodeSelector: {}
 tolerations: []
 affinity: {}
+
+# This is for setting up a PodDisruptionBudget more information can be found here:
+# https://kubernetes.io/docs/concepts/workloads/pods/disruptions/
+podDisruptionBudget: {}
+  # minAvailable: 1
+
+# Array of extra K8s manifests to deploy, e.g. an ExternalSecret for upstream mTLS credentials.
```
pseudomuto, line 136: /nit let's add a comment here about namespace/labels/etc. not being applied automatically like other templates.

charts/temporal-proxy/tests/extra_objects_test.yaml
```diff
@@ -0,0 +1,75 @@
...
+                property: password
+    asserts:
+      - containsDocument:
+          kind: ExternalSecret
+          apiVersion: external-secrets.io/v1beta1
+      - equal:
+          path: metadata.name
+          value: test-secret
+
+  - it: supports Go templating within entries
```
pseudomuto, line 61: Let's add one more test case that creates multiple extra objects, just to prove it works as expected.

pseudomuto (commented):
Hey @wankhede04, thanks for the PR!

I have a couple of minor comments below, but this LGTM 👍  Once addressed we'll get this merged.

charts/temporal-proxy/values.yaml
wankhede04: Good catch — added a note in values.yaml (fixed in 6e703a5) clarifying that extraObjects are rendered as-is and don't automatically inherit the release namespace, labels, or other chart-wide values.

charts/temporal-proxy/tests/extra_objects_test.yaml
wankhede04: Added a new test case "renders multiple extra objects" verifying two ConfigMaps render correctly from the array (fixed in 6e703a5).

wankhede04:
Hey @pseudomuto, I've addressed both review comments in 6e703a5:
- Added a note in `values.yaml` clarifying that `extraObjects` are rendered as-is and don't automatically inherit the release namespace, labels, or other chart-wide values.
- Added a new test case ("renders multiple extra objects") verifying multiple objects render correctly from the array.

All unit tests pass (44/44). If everything looks good now, could you please merge this PR? Thanks for the review!

pseudomuto (approved):
LGTM 👍 Thanks again for the PR @wankhede04!

PR #954, opened 2026-07-22
Tls wiring helpers

robholland:
## What was changed
Added `server.tls`, `web.tls` and `admintools.tls` stanzas. Also added a switch to use external frontend rather than the default of using the internal frontend for admintools/jobs (if enabled). Add some example values files and test TLS setups both with and without internal frontend. Based on the feature request and implementation by @luismacosta.

## Why?
This makes it easier for the user to configure TLS by avoiding boilerplate and ensures that TLS scenarios are tested in CI to avoid regressions. This makes it easy to integrate cert-manager via `extraObjects`.

## Checklist

1. Closes https://github.com/temporalio/helm-charts/pull/820

2. How was this tested:
Unit and integration tests added.

3. Any docs updates needed?
Docs updates included.


jiechenz (changes requested):
Claude found a bug when requireClientAuth is enabled through the following configuration:
```
  server:
    config:
      tls:
        frontend:
          server:
            requireClientAuth: true
```
The frontend correctly requires client authentication, but clients are missing the cert + key.

jiechenz (approved):
Some conflicts need to resolve, otherwsie lgtm

PR #960, opened 2026-07-28
feat(temporal-proxy): Add config model with secret and TLS wiring

pseudomuto:
The chart supplied config through a single `.Values.config` map rendered into a ConfigMap. That left no clean path for secrets (they landed in a plaintext ConfigMap), no first-class TLS/cert file handling, and a latent break: `tpl` ran over the config but the proxy uses the same `{{ }}` delimiters for its own per-request templates, so the official cloud example failed to render.

`.Values.config` now stays a faithful passthrough of the proxy schema, rendered by a _temporal-proxy.rendered_ helper rather than `tpl`, so proxy templates like `{{ .RemoteNamespace }}` pass through verbatim while hostPort still defaults to the service port. The few secret- and cert-bearing fields become chart-aware:

- `upstreams[].credentials.static.apiKey` and `auth.staticToken.token` accept a _secretKeyRef_; the chart injects a container env var (`TP_UPSTREAM_<NAME>_API_KEY` or `TP_AUTH_STATIC_TOKEN`) and substitutes `${VAR}` into the config, so the ConfigMap never holds a secret.
- gateway `config.tls` and each `upstreams[].tls` accept a _secretName_; the chart mounts that Secret at `/etc/temporal-proxy/certs/<slot>` and fills cert/key (and ca only when caKey is set) with the file paths. This works with cert-manager without requiring it. Upstream names that would produce an invalid DNS-1123 volume name fail the render up front.
- `env` / `envFrom` passthrough is added for values referenced with `${VAR}` in the config (e.g. `TEMPORAL_ACCOUNT`, or KMS credentials).

### DIFF
--- a/charts/temporal-proxy/README.md
+++ b/charts/temporal-proxy/README.md
@@ -31,5 +31,160 @@ helm install temporal-proxy ./charts/temporal-proxy
 ## Configuration
 
 The proxy's runtime configuration is set under `config` in `values.yaml` and rendered into a ConfigMap mounted at
-`/etc/temporal-proxy/config.yaml`. Values are treated as templates, so they can reference other values. See
-[`values.yaml`](./values.yaml) for the full set of options and their defaults.
+`/etc/temporal-proxy/config.yaml`. See [`values.yaml`](./values.yaml) for the full set of options and their defaults.
+
+## Supplying configuration
+
+### Inline config
+
+`config` mirrors the [temporal-proxy config schema](https://github.com/temporalio/temporal-proxy) directly; whatever you
+put under `config` in `values.yaml` is what ends up in the ConfigMap. The chart does not run this through Helm's `tpl`
+function, so the proxy's own per-request templates (for example `{{ .RemoteNamespace }}` in an upstream `hostPort`) pass
+through untouched and are evaluated by the proxy itself at request time, not by Helm.
+
+If `config.hostPort` (the gateway's listen address) is left unset, it defaults to `:<service.port>`.
+
+```yaml
+config:
+  routing:
+    default: default
+  upstreams:
+    - name: default
+      hostPort: localhost:7233
+```
+
+### Secret-aware credentials
+
+Two fields accept either a plain string or a `secretKeyRef` object, so the ConfigMap never has to hold a secret in
+plaintext:
+
+- `upstreams[].credentials.static.apiKey`
+- `auth.staticToken.token`
+
+When you use the `secretKeyRef` form, the chart adds an environment variable to the proxy container sourced from that
+Secret, and rewrites the config value to `${VAR}` so the proxy substitutes it at startup. The generated variable names
+are:
+
+- `TP_UPSTREAM_<NAME>_API_KEY` for `upstreams[].credentials.static.apiKey`, where `<NAME>` is the upstream's `name`
+  upper-cased (non-alphanumeric characters become `_`)
+- `TP_AUTH_STATIC_TOKEN` for `auth.staticToken.token`
+
+```yaml
+config:
+  upstreams:
+    - name: cloud
+      hostPort: localhost:7233
+      credentials:
+        static:
+          apiKey:
+            secretKeyRef:
+              name: temporal-cloud
+              key: api-key
+```
+
+renders a ConfigMap with `apiKey: ${TP_UPSTREAM_CLOUD_API_KEY}` and adds a matching `TP_UPSTREAM_CLOUD_API_KEY`
+environment variable to the container, backed by the `temporal-cloud`/`api-key` Secret.
+
+### TLS
+
+The gateway's `config.tls` block and each `upstreams[].tls` block accept a `secretName`. When set, the chart mounts that
+Secret at `/etc/temporal-proxy/certs/gateway` (for `config.tls`) or `/etc/temporal-proxy/certs/upstream-<name>` (for an
+upstream's `tls`), and sets:
+
+- `cert` to `<mount>/tls.crt` by default, or `<mount>/<certKey>` if `certKey` is set
+- `key` to `<mount>/tls.key` by default, or `<mount>/<keyKey>` if `keyKey` is set
+- `ca` only when `caKey` is set, to `<mount>/<caKey>`
+
+`ca` is opt-in: it is only written when `caKey` is provided. Setting a `ca` on the gateway's `config.tls` enables
+mTLS/client-cert enforcement for inbound connections, so only set it when you intend to require client certificates.
+
+Note: keep upstream `name` values DNS-safe (lowercase alphanumeric characters and `-`), since they are used to build the
+generated volume name (`certs-upstream-<name>`).
+
+This mounting scheme lines up with how [cert-manager](https://cert-manager.io/) issues Secrets (a `tls.crt`/`tls.key`
+pair, plus `ca.crt` when using `Certificate.spec.additionalOutputFormats` or a CA issuer), but cert-manager is not
+required; any Secret with the right keys works, and `certKey`/`keyKey`/`caKey` let you point at whatever keys your
+Secret actually uses.
+
+```yaml
+config:
+  tls:
+    secretName: temporal-proxy-server-tls
+  upstreams:
+    - name: cloud
+      hostPort: "{{ .RemoteNamespace }}.tmprl.cloud:7233"
+      tls:
+        secretName: temporal-cloud-client-tls
+        caKey: ca.crt
+```
+
+If you'd rather manage certificate files yourself, skip `secretName` and use the chart's generic `volumes` /
+`volumeMounts` values to mount them at whatever path you set `cert`/`key`/`ca` to.
+
+### env / envFrom passthrough
+
+`env` and `envFrom` on the Deployment let you supply arbitrary environment variables the config references with `${VAR}`
+syntax, such as a Temporal Cloud account id or KMS credentials. Chart-generated secret env vars (from the secret-aware
+fields above) are merged in automatically alongside anything you set in `.Values.env`.
+
+```yaml
+env:
+  - name: TEMPORAL_ACCOUNT
+    value: a1b2c
+
+envFrom:
+  - secretRef:
+      name: proxy-kms-credentials
+```
+
+### Full example: Temporal Cloud
+
+Putting the pieces above together, an upstream routing to Temporal Cloud with the API key sourced from a Secret and the
+namespace derived from the proxy's own per-request template:
+
+```yaml
+env:
+  - name: TEMPORAL_ACCOUNT
+    value: a1b2c
+  - name: TEMPORAL_NAMESPACE
+    value: testing
+
+extraObjects:
+  # Typically not created like this to avoid putting your API key in plain text.
+  # Demo purposes only.
+  - |
+    apiVersion: v1
+    kind: Secret
+    metadata:
+      name: temporal-cloud
+    type: Opaque
+    stringData:
+      api-key: <YOUR_API_KEY>
+
+config:
+  routing:
+    default: cloud
+    system: system
+  upstreams:
+    - name: cloud
+      hostPort: "{{ .RemoteNamespace }}.tmprl.cloud:7233"
+      namespaces:
+        rules:
+          suffix: .$TEMPORAL_ACCOUNT
+      tls: {}
+      credentials:
+        static:
+          apiKey:
+            secretKeyRef:
+              name: temporal-cloud
+              key: api-key
+    - name: system
+      hostPort: "$TEMPORAL_NAMESPACE.$TEMPORAL_ACCOUNT.tmprl.cloud:7233"
+      tls: {}
+      credentials:
+        static:
+          apiKey:
+            secretKeyRef:
+              name: temporal-cloud
+              key: api-key
+```
--- a/charts/temporal-proxy/templates/_helpers.tpl
+++ b/charts/temporal-proxy/templates/_helpers.tpl
@@ -71,3 +71,108 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+{{/*
+Rewrite a secret-bearing config field into a ${VAR} reference.
+
+If holder[field] is a { secretKeyRef: {...} } map, replace it with "${var}"
+and append a container env var (name=var, valueFrom.secretKeyRef=ref) to
+acc.env. A plain string value is left untouched. Any other map fails the
+render, since a near-miss key such as secretRef would otherwise land in the
+ConfigMap verbatim and only surface as an unmarshalling error at proxy
+startup. Mutates holder and acc in place and emits no output.
+
+Args (dict): holder, field, var, acc, path (config path, for error messages).
+*/}}
+{{- define "temporal-proxy.secretRef" -}}
+{{- $val := index .holder .field -}}
+{{- if kindIs "map" $val -}}
+{{- if not (hasKey $val "secretKeyRef") -}}
+{{- fail (printf "%s must be a string or a secretKeyRef" .path) -}}
+{{- end -}}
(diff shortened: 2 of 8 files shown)


robholland (commented):
Suggestions for the test suite, from reviewing the new config model. Split into three so they apply independently.

- The first (on `configmap_test.yaml`) adds **three failing tests** for input the model accepts today and arguably shouldn't. They're red until the helpers get matching `fail` guards, so treat them as bug reports rather than ready-to-merge.
- The other two add **nine passing tests** covering paths the new suite leaves unasserted.

Verified with `helm unittest charts/temporal-proxy` on the branch: 3 failed, 67 passed. With the prototype guards applied, 70/70 pass and the default render stays byte-identical to `main`.

Unrelated nit while I was in here: `it: templates config against other values` (line 38) no longer templates anything now that `tpl` is gone — it just checks the `service.port` default, which duplicates the new `defaults hostPort from the service port when unset`.

charts/temporal-proxy/tests/configmap_test.yaml
```diff
@@ -70,3 +70,142 @@ tests:
...
+        upstreams:
+          - name: Bad_Name
+            hostPort: cloud.example.com:7233
+            tls:
+              secretName: cloud-mtls
+        routing:
+          default: Bad_Name
+    asserts:
+      - failedTemplate:
+          errorMessage: "upstream name \"Bad_Name\" must be DNS-1123 safe (lowercase alphanumeric and '-') to mount a tls secret"
```
robholland: Three cases the config model accepts today that I think it should reject. Grouping them next to the DNS-1123 test since they're the same kind of guard.

**These are red as written** — they assert error messages the helpers don't emit yet, so they need the matching `fail` guards in `temporal-proxy.secretRef` / `temporal-proxy.tlsMount` before they'll pass. Worth reading as bug reports first; I've prototyped the guards and can send them if useful.

1. **`apiKey` as a map that isn't a `secretKeyRef`** — `secretRef` typo'd for `secretKeyRef` renders fine, the map lands in the ConfigMap, no env var is generated, and the proxy dies unmarshalling a map into `StaticCredentialConfig.APIKey`. Same hole on `auth.staticToken.token` via the shared helper.
2. **Nameless upstream + secret ref** — fails today with Helm's internal `wrong type for value; expected string; got interface {}` from `regexReplaceAll` at `_helpers.tpl:142`. Pre-PR this rendered and the proxy reported `upstreams[0].name is required`.
3. **`cert`/`key` alongside `secretName`** — `tlsMount` overwrites both unconditionally, so a user who asks for `/my/own/cert.pem` silently gets `/etc/temporal-proxy/certs/gateway/tls.crt`. I've asserted fail-on-conflict; honouring the explicit paths instead would be just as defensible, silence isn't.

The `FAILS TODAY` comments are triage aids — strip them once the guards land.

```suggestion
          errorMessage: "upstream name \"Bad_Name\" must be DNS-1123 safe (lowercase alphanumeric and '-') to mount a tls secret"

  # FAILS TODAY. temporal-proxy.secretRef only rewrites a map that has a
  # secretKeyRef key; any other map falls through into the ConfigMap verbatim.
  # A near-miss like `secretRef` therefore renders successfully, ships no env
  # var, and the proxy fails at startup unmarshalling a map into
  # StaticCredentialConfig.APIKey (a string). The same hole exists for
  # auth.staticToken.token, which shares the helper.
  - it: fails when an apiKey is a map that is not a secretKeyRef
    set:
      config:
        upstreams:
          - name: cloud
            hostPort: cloud.example.com:7233
            credentials:
              static:
                apiKey:
                  secretRef:
                    name: temporal-cloud
                    key: api-key
        routing:
          default: cloud
    asserts:
      - failedTemplate:
          errorMessage: "upstreams[cloud].credentials.static.apiKey must be a string or a secretKeyRef"

  # FAILS TODAY, with a Helm-internal message rather than a useful one:
  #   wrong type for value; expected string; got interface {}
  # A nameless upstream reaches regexReplaceAll with a nil name while building
  # the env var. The DNS-1123 guard covers this for the tls path but not the
  # credentials path. Before this PR the config rendered and the proxy reported
  # `upstreams[0].name is required`, so the error quality regressed.
  - it: fails clearly when an upstream with a secret ref has no name
    set:
      config:
        upstreams:
          - hostPort: cloud.example.com:7233
            credentials:
              static:
                apiKey:
                  secretKeyRef:
                    name: temporal-cloud
                    key: api-key
        routing:
          default: cloud
    asserts:
      - failedTemplate:
          errorMessage: "upstreams[0] must have a name to wire its credentials or tls secret"

  # FAILS TODAY. tlsMount overwrites cert/key unconditionally, so explicit
  # paths supplied alongside secretName are silently discarded -- the user asks
  # for /my/own/cert.pem and gets /etc/temporal-proxy/certs/gateway/tls.crt.
  # Failing on the conflict is asserted here; honouring the explicit paths
  # instead would be equally defensible. Silently ignoring them is not.
  - it: fails when a tls block sets both secretName and an explicit cert
    set:
      config:
        tls:
          secretName: proxy-gateway-tls
          cert: /my/own/cert.pem
          key: /my/own/key.pem
        upstreams:
          - name: cloud
            hostPort: cloud.example.com:7233
        routing:
          default: cloud
    asserts:
      - failedTemplate:
          errorMessage: "tls for gateway sets both secretName and cert/key; use certKey/keyKey to point at other keys in the Secret, or drop secretName and mount the files yourself"
```

charts/temporal-proxy/tests/configmap_test.yaml
```diff
@@ -70,3 +70,142 @@ tests:
...
+                apiKey:
+                  secretKeyRef:
+                    name: temporal-cloud
+                    key: api-key
+        routing:
+          default: My_Upstream
+    asserts:
+      - matchRegex:
+          path: data["config.yaml"]
+          pattern: "apiKey: \\$\\{TP_UPSTREAM_MY_UPSTREAM_API_KEY\\}"
```
robholland, line 369: Five gaps in the config-side coverage. **These all pass against the branch as-is** — they pin behaviour that currently works but isn't asserted anywhere:

- a literal `auth.staticToken.token` left untouched (`apiKey` has both a secret-ref and a literal test, the token only has the secret-ref one)
- `certKey` / `keyKey` overrides — the documented escape hatch for Secrets that don't use the cert-manager layout, currently untested
- `caKey` on an *upstream* slot, which is the more common place for it than the gateway
- chart-only keys stripped from an *upstream* `tls` (only the gateway asserts this, and leaking them would make the proxy reject the config)
- `serverName` surviving `tlsMount`, which `unset`s four keys and rewrites three more

```suggestion
          pattern: "apiKey: \\$\\{TP_UPSTREAM_MY_UPSTREAM_API_KEY\\}"

  # Coverage gap: apiKey has both a secret-ref and a literal case, the static
  # token only has the secret-ref one. Pins that a plain string token is left
  # alone and no env var substitution leaks into it.
  - it: leaves a literal static token string untouched
    set:
      config:
        auth:
          staticToken:
            token: "$MY_OWN_TOKEN_VAR"
            header: Authorization
    asserts:
      - matchRegex:
          path: data["config.yaml"]
          pattern: "token: \\$MY_OWN_TOKEN_VAR"
      - notMatchRegex:
          path: data["config.yaml"]
          pattern: "TP_AUTH_STATIC_TOKEN"

  # Coverage gap: only the caKey default is exercised. certKey/keyKey are the
  # documented escape hatch for Secrets that do not use the cert-manager
  # tls.crt/tls.key layout.
  - it: honours certKey and keyKey overrides
    set:
      config:
        tls:
          secretName: proxy-gateway-tls
          certKey: server.pem
          keyKey: server-key.pem
        upstreams:
          - name: cloud
            hostPort: cloud.example.com:7233
        routing:
          default: cloud
    asserts:
      - matchRegex:
          path: data["config.yaml"]
          pattern: "cert: /etc/temporal-proxy/certs/gateway/server.pem"
      - matchRegex:
          path: data["config.yaml"]
          pattern: "key: /etc/temporal-proxy/certs/gateway/server-key.pem"

  # Coverage gap: caKey is only tested on the gateway slot. An upstream is the
  # more common place for it (verifying a self-hosted cluster's private CA).
  - it: sets the ca path on an upstream tls when a caKey is provided
    set:
      config:
        upstreams:
          - name: onprem
            hostPort: temporal-frontend.temporal.svc:7233
            tls:
              secretName: onprem-client-tls
              caKey: ca.crt
        routing:
          default: onprem
    asserts:
      - matchRegex:
          path: data["config.yaml"]
          pattern: "ca: /etc/temporal-proxy/certs/upstream-onprem/ca.crt"
      - matchRegex:
          path: data["config.yaml"]
          pattern: "cert: /etc/temporal-proxy/certs/upstream-onprem/tls.crt"

  # Coverage gap: the gateway test asserts the chart-only keys are stripped,
  # the upstream path is unverified. Leaking them would make the proxy reject
  # the config, since TLSConfig has no such fields.
  - it: strips chart-only tls keys from an upstream tls block
    set:
      config:
        upstreams:
          - name: onprem
            hostPort: temporal-frontend.temporal.svc:7233
            tls:
              secretName: onprem-client-tls
              certKey: server.pem
              keyKey: server-key.pem
              caKey: ca.crt
        routing:
          default: onprem
    asserts:
      - notMatchRegex:
          path: data["config.yaml"]
          pattern: "secretName"
      - notMatchRegex:
          path: data["config.yaml"]
          pattern: "certKey"
      - notMatchRegex:
          path: data["config.yaml"]
          pattern: "keyKey"
      - notMatchRegex:
          path: data["config.yaml"]
          pattern: "caKey"

  # Coverage gap: tlsMount unsets four keys by name and rewrites three more.
  # Nothing pins that the tls fields the chart does not manage survive it.
  - it: preserves tls fields the chart does not manage
    set:
      config:
        tls:
          secretName: proxy-gateway-tls
          serverName: proxy.example.com
        upstreams:
          - name: cloud
            hostPort: cloud.example.com:7233
        routing:
          default: cloud
    asserts:
      - matchRegex:
          path: data["config.yaml"]
          pattern: "serverName: proxy.example.com"
```

charts/temporal-proxy/tests/deployment_test.yaml
```diff
@@ -153,3 +153,175 @@ tests:
...
+          value: certs-gateway
+      - equal:
+          path: spec.template.spec.volumes[2].name
+          value: certs-upstream-cloud
+      - equal:
+          path: spec.template.spec.containers[0].volumeMounts[1].name
+          value: certs-gateway
+      - equal:
+          path: spec.template.spec.containers[0].volumeMounts[2].name
+          value: certs-upstream-cloud
```
robholland, line 327: Two more that **pass as-is**:

- **env ordering** — generated vars are emitted before `.Values.env`, which is what makes a user override effective (a duplicate env name is last-one-wins at the runtime). Swapping the two `with` blocks would silently make `.Values.env` a no-op, so it's worth pinning alongside the volume-ordering test above.
- **no stray mounts** — a config whose `tls: {}` blocks carry no `secretName` should still mount only the config volume.

```suggestion
          value: certs-upstream-cloud

  # Coverage gap: the chart emits generated env vars before .Values.env, which
  # is what lets a user override one -- a duplicate env name is resolved
  # last-one-wins by the runtime. Reversing the two blocks would silently make
  # .Values.env ineffective, so pin the order.
  - it: emits generated secret env vars before .Values.env
    set:
      env:
        - name: TP_UPSTREAM_CLOUD_API_KEY
          value: my-own-literal
      config:
        upstreams:
          - name: cloud
            hostPort: cloud.example.com:7233
            credentials:
              static:
                apiKey:
                  secretKeyRef:
                    name: temporal-cloud
                    key: api-key
        routing:
          default: cloud
    asserts:
      # index 0 is PROXY_CONFIG, 1 the generated var, 2 the user's override.
      - equal:
          path: spec.template.spec.containers[0].env[1].name
          value: TP_UPSTREAM_CLOUD_API_KEY
      - equal:
          path: spec.template.spec.containers[0].env[1].valueFrom.secretKeyRef.name
          value: temporal-cloud
      - equal:
          path: spec.template.spec.containers[0].env[2].name
          value: TP_UPSTREAM_CLOUD_API_KEY
      - equal:
          path: spec.template.spec.containers[0].env[2].value
          value: my-own-literal

  # Coverage gap: nothing asserts that a config with no secretName mounts
  # nothing extra -- the default path should carry only the config volume.
  - it: mounts only the config volume when no tls secret is set
    set:
      config:
        tls: {}
        upstreams:
          - name: cloud
            hostPort: cloud.example.com:7233
            tls: {}
        routing:
          default: cloud
    asserts:
      - lengthEqual:
          path: spec.template.spec.volumes
          count: 1
      - lengthEqual:
          path: spec.template.spec.containers[0].volumeMounts
          count: 1
```

charts/temporal-proxy/README.md
```diff
@@ -31,5 +31,144 @@ helm install temporal-proxy ./charts/temporal-proxy
...
+          suffix: .$TEMPORAL_ACCOUNT
+      tls: {}
+      credentials:
+        static:
+          apiKey:
+            secretKeyRef:
+              name: temporal-cloud
+              key: api-key
+    - name: system
+      hostPort: "$TEMPORAL_NAMESPACE.$TEMPORAL_ACCOUNT.tmprl.cloud:7233"
```
robholland, line 182: `TEMPORAL_NAMESPACE` isn't set in the `env` map.

pseudomuto:
Hey @robholland, thanks for the solid review!

I've incorporated your suggestions and feedback and updated the PR accordingly. When you get a chance, PTAL.

charts/temporal-proxy/README.md
```diff
@@ -31,5 +31,159 @@ helm install temporal-proxy ./charts/temporal-proxy
...
+env:
+  - name: TEMPORAL_ACCOUNT
+    value: a1b2c
+  - name: TEMPORAL_NAMESPACE
+    value: testing
+
+extraObjects:
+  # Typically not created like this to avoid putting your API key in plain text.
+  # Demo purposes only.
+  - apiVersion: v1
```
robholland: Either the extraObjects template needs to also support maps here (but only `tpl` the strings, I guess?), or this needs to be a string.

charts/temporal-proxy/templates/_helpers.tpl
```diff
@@ -71,3 +71,105 @@ Create the name of the service account to use
...
+{{- define "temporal-proxy.tlsMount" -}}
+{{- $tls := .tls -}}
+{{- if or (hasKey $tls "cert") (hasKey $tls "key") -}}
+{{- fail (printf "tls for %s sets both secretName and cert/key; use certKey/keyKey to point at other keys in the Secret, or drop secretName and mount the files yourself" .slot) -}}
+{{- end -}}
+{{- $mount := printf "/etc/temporal-proxy/certs/%s" .slot -}}
+{{- $name := printf "certs-%s" .slot -}}
+{{- $_ := set $tls "cert" (printf "%s/%s" $mount (default "tls.crt" $tls.certKey)) -}}
+{{- $_ := set $tls "key" (printf "%s/%s" $mount (default "tls.key" $tls.keyKey)) -}}
+{{- if $tls.caKey -}}
```
robholland, line 123: I think we should `fail` here if `ca` and `caKey` are both set as `ca` will be ignored.

robholland (approved):
Two small comments, otherwise lgtm.

Does the proxy support credential reloading? There is no mention in the Helm README about how creds/certs can be rotated. The Deployment template doesn't use any config checksums or similar, so afaik pods will not be restarted when/if config or secrets change (and Helm doesn't write the secrets anyway so can't use checksum there).

charts/temporal-proxy/README.md
pseudomuto: This is definitely a typo 🤦 I meant to have `- |` in there. I'll update.

pseudomuto:
> Does the proxy support credential reloading?

Currently not. Most connections (unless templated hosts) are established at startup using the credentials identified in the config. If they change/expire, errors will go up and to the right.

PR #961, opened 2026-07-29
Default shims off

### LINKED ISSUE #940: [Feature Request] set shims.dockerize and shims.elasticsearchTool default to false in upcoming Chart Releases

The chart ships with:

shims:
  dockerize: true
  elasticsearchTool: true

The comments on those same values say:

▎ Enable compatibility with Temporal 1.29 images. Set to false if using Temporal 1.30 or higher.

The appVersion in Chart.yaml is 1.31.0 and the default image tags are 1.31.0, so the defaults are not sensible.

Flip both shims defaults to false. Two-line change.

This eliminates the surprise a user experiences on an install using defaults: the current defaults create a ConfigMap, volumes, and volumeMounts in every server pod that do nothing. Setting both to false removes those resources and reduces the rendered manifest by 364 lines (1,556 → 1,192) and improves the supportability of a default install.

### Additional context

Confirmed behavior (tested on chart 1.2.0 / Temporal 1.31.0 on minikube):

- The temporal-shims ConfigMap is created and the fake dockerize script is volume-mounted at /usr/local/bin/dockerize in every server pod
- The 1.31.0 image's entrypoint.sh runs exec temporal-server start — it never calls dockerize
- No rendered config.yaml is produced; temporal-server processes config_template.yaml internally using its own Go template engine
- The shim is mounted but never executed

This issue was initiated by a user and AI (Claude) was used to assist with testing and drafting this bug report.

robholland:
## What was changed
Default the `dockerize` and `elasticsearch-tool` shims off now.

## Why?
These haven't been needed for the last two major Temporal releases we've shipped and can cause confusion for users.

## Checklist

1. Closes #940.

2. How was this tested:
Tests updated.

### DIFF
--- a/charts/temporal/values.yaml
+++ b/charts/temporal/values.yaml
@@ -657,9 +657,9 @@ schema:
   securityContext: {}
 shims:
   # Enable compatibility with Temporal 1.29 images. Set to false if using Temporal 1.30 or higher.
-  dockerize: true
+  dockerize: false
   # Enable compatibility with Temporal 1.29 images. Set to false if using Temporal 1.30 or higher.
-  elasticsearchTool: true
+  elasticsearchTool: false
 test:
   podAnnotations: {}
   podLabels: {}
--- a/charts/temporal/values.yaml
+++ b/charts/temporal/values.yaml
@@ -656,9 +656,9 @@ schema:
   containerSecurityContext: {}
   securityContext: {}
 shims:
-  # Enable compatibility with Temporal 1.29 images. Set to false if using Temporal 1.30 or higher.
+  # Enable compatibility with Temporal 1.29 images. Set to true if using Temporal 1.29.
   dockerize: false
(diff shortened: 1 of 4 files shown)


charts/temporal/values.yaml
```diff
@@ -656,10 +656,10 @@ schema:
   containerSecurityContext: {}
   securityContext: {}
 shims:
-  # Enable compatibility with Temporal 1.29 images. Set to false if using Temporal 1.30 or higher.
-  dockerize: true
-  # Enable compatibility with Temporal 1.29 images. Set to false if using Temporal 1.30 or higher.
-  elasticsearchTool: true
+  # Enable compatibility with Temporal 1.29 images. Set to true if using Temporal 1.29.
+  dockerize: false
+  # Enable compatibility with Temporal 1.29 images. Set to true if using Temporal 1.29.
```
pseudomuto, line 661: /nit Should this be "Temporal < 1.30`?

pseudomuto (approved):

PR #982, opened 2026-09-15
fix(temporal-proxy): probe the gRPC health service instead of httpGet

Vaughan-Temporal:
## Summary

The `temporal-proxy` chart shipped the `helm create` scaffold probes unchanged:

```yaml
livenessProbe:
  httpGet: { path: /, port: http }
readinessProbe:
  httpGet: { path: /, port: http }
```

The proxy serves gRPC on that port and has no HTTP handler on it, so **neither probe could ever pass**. The pod never went Ready, so its Service had no endpoints and the proxy was unreachable through it, while liveness restarted the container in a loop. A rollout with `maxUnavailable: 0` could never progress.

The kubelet's prober speaks HTTP/1.1 and gets an HTTP/2 SETTINGS frame back; forced onto HTTP/2 it gets `415 invalid gRPC request content-type`. Both are probe failures.

A second bug was masking the first: the default `config` pointed at a plaintext `localhost:7233` with no `insecure`, so the proxy TLS-dialled it, failed `WaitReady` and exited before binding the gateway. `helm install` with stock values gave CrashLoopBackOff before the probes mattered.

## Changes

Probes now carry timings only, and the chart supplies the handler:

| Gateway | Handler | Why |
| --- | --- | --- |
| plaintext (default) | `grpc` | Checks the health service the proxy already serves. Needs Kubernetes 1.27. |
| TLS (`config.tls` set) | `tcpSocket` | The kubelet's `grpc` probe dials plaintext and cannot do TLS ([kubernetes/enhancements#4939](https://github.com/kubernetes/enhancements/issues/4939)). |

Without that split the fix only moves the failure — a `grpc` probe against a TLS gateway times out and restarts the container just the same. The `temporal` chart hit the same limitation and made the same choice.

Handlers are left out of the defaults on purpose: Helm merges values rather than replacing them, so a default handler would survive alongside one a user sets and produce a two-handler probe the API server rejects. A `grpc` probe takes a numeric port and can't name a container port, so an unset one is filled in from `service.port`.

Liveness stays local-only — a restart fixes process-local faults, not an unreachable upstream, and gating it on upstream health would turn one blip into a fleet-wide restart. `timeoutSeconds` is raised off the 1s default, which is shorter than a GC pause.

No `startupProbe` is shipped: the proxy binds the gateway only once its upstreams are ready, and that is bounded by the fx start timeout (15s), which `failureThreshold: 4` at `periodSeconds: 15` already tolerates. The passthrough is kept for when `livenessProbe` is tightened below that ceiling, and the README says when.

Also adds `insecure: true` to the default upstream so it boots. Helm replaces arrays, so it can't leak into a user-supplied `upstreams` list.

## Testing

- **`origin/main`:** `0/1 Running`, restarts climbing, `Liveness probe failed: ... malformed HTTP response "\x00\x00\x06\x04..."` → `failed liveness probe, will be restarted`.
- **This PR, plaintext:** `1/1`, 0 restarts, 0 `Unhealthy` events; `grpc.health.v1.Health/Check` returns `SERVING` through the Service and `temporal operator cluster system` reaches the upstream.
- **This PR, TLS gateway:** `1/1`, 0 restarts, 0 `Unhealthy` events.
- **Counterfactual:** forcing `grpc` onto the TLS gateway fails — `timeout: failed to connect service "10.244.0.8:8443" within 5s`, container restarted. The `tcpSocket` fallback is required, not defensive.

## Compatibility

Non-breaking; chart `0.2.1` → `0.2.2`. No template renames, no Service changes.

Since the default probes never passed, no deployment can have been running successfully on them. The affected population is users who already overrode the probes:

- **Overrode with another handler:** previously merged into a two-handler probe the API server rejects, so those installs already failed unless `httpGet` was also nulled. Now valid, carrying the chart's timings.
- **Nulled both probes:** unaffected, and pinned by a test so a future `startupProbe` default can't silently reappear for them.
- **Clusters older than 1.27:** the `grpc` handler needs 1.27 (beta and on by default from 1.24). The README documents the `grpc-health-probe` `exec` probe, which also supports TLS. No `kubeVersion` gate was added, since a TLS gateway renders `tcpSocket` and doesn't need 1.27.


charts/temporal-proxy/Chart.yaml
```diff
@@ -4,7 +4,7 @@ description: The Temporal proxy handles routing between clusters, namespace tran
 icon: https://raw.githubusercontent.com/temporalio/temporal-proxy/refs/heads/main/temporal-proxy-small.png
 
 type: application
-version: 0.2.1
+version: 0.2.2
 appVersion: v0.5.0
```
pseudomuto, line 8: Can we bump to v0.7.0?

pseudomuto (approved):

